PR #3, opened 2016-01-31
tab left

JeremyLi28:
Just tab list column. 
TODO:
tab right column
make them scrollable


JavierJia:
@JeremyLi28 Could you give me an screenshot about the effect of tablized UI?

JeremyLi28:
sure

PR #11, opened 2016-04-09
Using JSONP

JeremyLi28:
New branch

Prettify the tweet table done.
There are still some problems:
1. It seems that the normal approach is letting the server sends the request to twitter api not the browser, since using the browser will suffer from the cross domain problem
2. There are 2 approaches to solve the cross domain problem, CORS and JSONP, CORS is more recommended and JSONP is a little old, but it seems that twitter api not support CORS, so I use JSONP.
3. I test in Safari, Chrom and Firefox, some extension may affect the display in Safari, others are all good.
4. Still working on the error handling of JSONP, it's a little tricky, currently it will report 404 if the tweets can not be found( which is normal) in the console

### DIFF
--- a/src/main/resources/twitter-demo/static/js/tweetbook.js
+++ b/src/main/resources/twitter-demo/static/js/tweetbook.js
@@ -454,7 +454,7 @@ function buildTweetSample(type, parameters) {
     aql.push('where $t.create_at >= $ts_start and $t.create_at < $ts_end');
   }
   aql.push('limit 100');
-  aql.push('return {"uname": $t.user.screen_name, "tweet":$t.text_msg};\n')
+  aql.push('return {"uname": $t.user.screen_name, "tweet":$t.text_msg, "id":$t.id};\n')
   return aql.join('\n');
 }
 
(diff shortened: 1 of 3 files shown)


src/main/resources/twitter-demo/static/js/tweetbook.js
```diff
@@ -755,7 +755,15 @@ function drawHashtag(tag_count) {
  **/
 function drawTweets(message) {
   $.each(message, function (i, d) {
-    $('#tweets tr:last').after('<tr><td>' + d.uname + '<br/>' + d.tweet + '</td></tr>');
+    // Using JSONP
+    var url = "https://api.twitter.com/1/statuses/oembed.json?id=" + d.id;
+    $.ajax({
```
JavierJia: We should move this logic to the server part , i.e. run_demo.py script. Then frontend only need to request to the server of itself.

JeremyLi28: Sure, I'll do that.

JeremyLi28: Shall we store the request in AsterixDB? Since I saw someone stores all the html in the database. Or we just store it as a file?

JavierJia: Yes, we need to store the queries. However, I think the middleware is a good place to record such things.

JeremyLi28: This one is the prettify tweet branch

JavierJia: Got it. It seems not hard. We should put this request into the server(that python script), and only send the $.ajax to our own server.

JeremyLi28: Sure, let me have a try

JeremyLi28: Done moving query to server

JavierJia:
There is conflict with master, can you pull master and resolve the conflicts?

JeremyLi28:
Sure

JavierJia:
Great, let's merge it.

PR #12, opened 2016-04-11
sider bar vertical tab and click popup

JeremyLi28:
Make side bar tab vertical, and click popup. 
screen shot 2016-04-10 at 8 48 54 pm
screen shot 2016-04-10 at 8 49 03 pm


JavierJia:
Very nice! :+1:

PR #13, opened 2016-04-12
minor bug fix

JeremyLi28:
When zoom too far, the result will return empty list, I made some modification to ensure all the graph won't change if it is the case.


JavierJia:
great. :+1:

PR #21, opened 2016-04-27
Modularize the angular js

JavierJia:
@JeremyLi28 @xixZ


neo/app/assets/javascripts/common/services.js
```diff
@@ -0,0 +1,69 @@
+angular.module('cloudberry.common', [])
+  .service('Asterix', function($http, $timeout) {
```
JeremyLi28: Changing from factory to service?

JeremyLi28:
I think probably using angular to modularize is enough? It also has the dependency and stuff

neo/app/assets/javascripts/common/services.js
JavierJia: I'm not very clear about the difference between these two. In most of the cases, the service seems used more often for the shared utility functions. So I choose to make it as a service.
Do you know any drawback of using service?

JavierJia:
Absolutely! Angular itself is already modularized. I gave up the require js.

neo/app/assets/javascripts/common/services.js
JeremyLi28: No clear either... my understanding is service return a function and factory return a object which contains functions. But I think both works ok for our application

xixZ:
LGTM.

Issue #24, opened 2016-04-27
Create a NewYorkGnosis object to resolve the geographic queries


JavierJia:
We have the `USGeoGnosis` in the `gnosis` module. It is used by `noah` and `neo` for the purposes of geotagging data source and geotagging the box area from the query. 
We need a similar component and functionality to handle the NewYork data.

JavierJia:
@xixZ Could you also tag the pick-up position? we will need both tagged information.

xixZ:
Already tagged.

JavierJia:
nice.

PR #26, opened 2016-04-29
New York geoTag

xixZ:
@JavierJia @JeremyLi28 
#24


gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/IGeoIndex.scala
```diff
@@ -41,4 +41,32 @@ class USGeoJSONIndex() extends IGeoIndex {
   override def search(envelope: Envelope): Seq[IUSGeoJSONEntity] = {
     index.query(envelope).asScala.map(item => entities(item.asInstanceOf[Int]))
   }
+}
+
+class NYGeoJSONIndex() extends IGeoIndex {
```
JavierJia: what's the difference of this `NYGeoJsonIndex`? It seems should be an instantiated object of `USGeoJsonIndex` other than a separate class.

gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/NewYorkGnosis.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  lazy val neighborShapes: IGeoIndex = levelShapeMap.get(NeighborLevel).get
+
+  // used in geo tag
+  def tagPoint(longitude: Double, latitude: Double): Option[NYGeoTagInfo] = {
+    val box = new Envelope(new Coordinate(longitude, latitude))
+    neighborShapes.search(box).headOption.map(entity => NYGeoTagInfo(entity.asInstanceOf[NYNeighborEntity]))
+  }
+
```
JavierJia: we also need a similar `tagRectangle` method to answer the query from the frontend which give us a rectangle and answer it with a list of neighbors/boro.

gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/NewYorkGnosis.scala
```diff
@@ -0,0 +1,66 @@
+package edu.uci.ics.cloudberry.gnosis
+
+import java.io.{File, FilenameFilter}
+
+import com.vividsolutions.jts.geom.{Coordinate, Envelope, Geometry}
+import play.api.libs.json.{JsObject, Json, Writes}
+
+class NewYorkGnosis(levelGeoPathMap: Map[TypeLevel, File]) {
```
JavierJia: Similar as previous `NYGeoIndex`, could you extract the common requirement from the both `gnosis` as some `Abstract` class, and let both concrete class extend on it?

JavierJia:
Just a few comments. Looks a good progress of learning Scala. 👍

gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/IGeoIndex.scala
xixZ: It can't be an instantiated object of USGeoJsonIndex, because they use different entities.

xixZ: Failed to find a way out. I tried to move loadShape to IGeoIndex and change IUSGeoJSONEntity into IEntity. Declare entities in IGeoIndex as ArrayBuffer[IEntity].
It can't compile successfully. @JavierJia Help!!!

JavierJia: what is the compilation error if you run `sbt compile`?

xixZ: In USGeoJsonIndex, lazy val entities: ArrayBuffer[IUSGeoJSONEntity] has incompatible type

JavierJia: I think you may also need to change the `implicit builder: (Map[String, AnyRef], Geometry) => INYGeoJSONEntity` to `implicit builder: (Map[String, AnyRef], Geometry) => IEntity`

gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/NewYorkGnosis.scala
```diff
@@ -5,13 +5,13 @@ import java.io.{File, FilenameFilter}
 import com.vividsolutions.jts.geom.{Coordinate, Envelope, Geometry}
 import play.api.libs.json.{JsObject, Json, Writes}
 
-class NewYorkGnosis(levelGeoPathMap: Map[TypeLevel, File]) {
+class NewYorkGnosis(levelGeoPathMap: Map[TypeLevel, File]) extends IGnosis{
 
   import NewYorkGnosis._
 
-  val levelShapeMap: Map[TypeLevel, NYGeoJSONIndex] = load(levelGeoPathMap)
+  lazy val levelShapeMap = load(levelGeoPathMap)
```
JavierJia: It's better to be not lazy, because when we start the server, we want to load everything and then begin to response. Otherwise the first query will be too slow.

gnosis/src/main/scala/edu/uci/ics/cloudberry/gnosis/USGeoGnosis.scala
```diff
@@ -7,13 +7,13 @@ import play.api.libs.json.{JsObject, Json, Writes}
 
 import scala.collection.mutable
 
-class USGeoGnosis(levelGeoPathMap: Map[TypeLevel, File]) {
+class USGeoGnosis(levelGeoPathMap: Map[TypeLevel, File]) extends IGnosis{
 
   import USGeoGnosis._
 
-  val levelShapeMap: Map[TypeLevel, USGeoJSONIndex] = load(levelGeoPathMap)
+  lazy val levelShapeMap: Map[TypeLevel, USGeoJSONIndex] = load(levelGeoPathMap)
```
JavierJia: ditto

xixZ:
@JavierJia

Issue #23, opened 2016-04-27
Move the UI components from the current cloudberry ui to play branch


JavierJia:
Including
- [x] Map
- [x] Timeline
- [x] HashTag
- [x] Actual Tweets.

JavierJia:
@JeremyLi28 could you list the rest of the work mentioned in the #27  to here? thank you!

JeremyLi28:
TODO:
1. Add the actual tweets.
2. Further Modulize for sidebar
3. Some styling issues
4. (Long term) Develop for general purpose.

PR #27, opened 2016-05-01
add timeseries directive

JeremyLi28:
rt


neo/app/assets/javascripts/app.js
```diff
@@ -17,7 +17,8 @@ app.controller('SearchCtrl', function($scope, $window, Asterix) {
       Asterix.parameters.time = $scope.time;
     if ($scope.level)
       Asterix.parameters.level = $scope.level;
-    Asterix.query(Asterix.parameters);
+    Asterix.queryType = 'search';
```
JavierJia: fix the indent?

JeremyLi28: which one? I don't see

neo/app/assets/javascripts/timeseries/controllers.js
```diff
@@ -0,0 +1,94 @@
...
+    );
+  })
+  .directive('timeSeries', function (Asterix) {
+    var margin = {
+      top: 10,
+      right: 10,
+      bottom: 30,
+      left: 50
+    };
+    var width = 962 - margin.left - margin.right;
```
JavierJia: could you replace these constant value `962` and `150` with some relative values?
I know it will look beautiful in the normal screen. However, we need to show the screen on the lower resolution projector, which should also look good.

JeremyLi28: Sure, I see this problem the first time I write it but couldn't find a good solution at that time, let have a try now

project/dependencies.scala
```diff
@@ -25,7 +25,10 @@ object Dependencies {
     "org.webjars" % "leaflet" % "0.7.7",
     "org.webjars" % "angular-leaflet-directive" % "0.8.2",
     "com.github.nscala-time" %% "nscala-time" % "2.10.0",
-    "org.scalactic" %% "scalactic" % "2.2.6"
+    "org.scalactic" %% "scalactic" % "2.2.6",
```
JavierJia: minor comment, could you move those `webjars` next to each other that looks like a group?

JeremyLi28: Sure

neo/app/assets/javascripts/app.js
JavierJia: my bad. never mind.

JavierJia:
@JeremyLi28 could you upload a sample picture to show the final effect?

JeremyLi28:
@JavierJia Will do it after I finish them all

JavierJia:
nice. Looking forward to the beautiful UI and the clean code. 😄

JeremyLi28:
@JavierJia It almost looks like before right now, except for the tweets, need some modification in the backend.
Also, haven't figure out how to build the sidebar directive.
Still some problem remaining, like the angular-leaflet package in webJar is not the most recently, so it can't set init zoom, it default zoom to 1 right now.
screen shot 2016-05-02 at 6 02 40 pm

JavierJia:
@JeremyLi28  looks great. Could you comment on the #23 issue to describe the rest of the work as a TODO list? we can finished them in the separate PR.

PR #47, opened 2016-05-25
fix the get tweet json api

JavierJia:
@JeremyLi28 fix the server to get sample tweets from twitter api.

### DIFF
--- a/neo/app/controllers/Application.scala
+++ b/neo/app/controllers/Application.scala
@@ -59,9 +59,11 @@ class Application @Inject()(val wsClient: WSClient,
     ActorFlow.actorRef(out => UserActor.props(out, cachesActor, USGeoGnosis))
   }
 
-  def tweet(id:String) = Action {
+  def tweet(id:String) = Action.async {
     val url = "https://api.twitter.com/1/statuses/oembed.json?id="+id
-    Redirect(url, 200)
+    wsClient.url(url).get().map { response =>
+      Ok(response.json)
+    }
   }
 
   def search(query: JsValue) = Action.async {


neo/app/controllers/Application.scala
```diff
@@ -59,9 +59,11 @@ class Application @Inject()(val wsClient: WSClient,
     ActorFlow.actorRef(out => UserActor.props(out, cachesActor, USGeoGnosis))
   }
 
-  def tweet(id:String) = Action {
+  def tweet(id:String) = Action.async {
```
chenlica: I am new to Scala, and can give superficial comments: can we add comments to the code? :-)

JavierJia: I'm very happy to explain Scala code in review comment, but I really don't like comments in the code which usually shows that the code is not clear enough 😢 

This function actually is a Play framework function which sends a request to that composed `url`. `wsClient.url().get` get the response back by return a `Future[WSResponse]`. Then we just wrap up the body of the response with a "OK 200" return code and send back to the client.

chenlica: Never mind then :-)

JavierJia: It's actually a pretty good practice to learn the code through code review. @JeremyLi28 @xixZ  you are very encouraged to ask questions about the unknown codes 😄

JeremyLi28:
So I can directly call the '/tweet/' route now?

JavierJia:
Yes. I tested it by simply open `localhosts:9000/tweet/715635459397967872`.

JeremyLi28:
LGTM

PR #49, opened 2016-05-25
Try using jsonp in frontend instead of asking server

JavierJia:
@JeremyLi28 
I just give a try, it seems working 😄

### DIFF
--- a/neo/app/assets/javascripts/sidebar/controllers.js
+++ b/neo/app/assets/javascripts/sidebar/controllers.js
@@ -29,16 +29,9 @@ angular.module('cloudberry.sidebar', ['cloudberry.common'])
     function drawTweets(message) {
       $('#tweet').html('');
       $.each(message, function (i, d) {
-        var url = "/tweet/" + d.tid;
-        $.ajax({
-          url: url,
-          dataType: "json",
-          success: function (data) {
-            $('#tweet').append(data.html);
-          },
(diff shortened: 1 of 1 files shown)


neo/app/assets/javascripts/sidebar/controllers.js
```diff
@@ -29,16 +29,9 @@ angular.module('cloudberry.sidebar', ['cloudberry.common'])
...
-        $.ajax({
-          url: url,
-          dataType: "json",
-          success: function (data) {
-            $('#tweet').append(data.html);
-          },
-          error: function(data) {
-            // do nothing
-          }
+        var url = "https://api.twitter.com/1/statuses/oembed.json?callback=JSON_CALLBACK&id=" + d.tid
```
JeremyLi28: what is the callback for? I think this is the only difference between our approach

JeremyLi28:
Cool! Let me check out

neo/app/assets/javascripts/sidebar/controllers.js
JavierJia: I don't know the details either. The [documentation](https://docs.angularjs.org/api/ng/service/$http#jsonp) says it has to be this name

JeremyLi28: Got it, thanks!

Issue #81, opened 2016-06-12
Handle multiple keywords 


JavierJia:
- [x] multiple keywords
- [x] none ASCII keyword

PR #84, opened 2016-06-15
Support the Non-ASCII UTF8 keywords

JavierJia:
This PR 

- Support the Non-ASCII UTF8 keywords. 
- Fix the front-end query value, so it will send the multi-keywords to the back-end.

Solving issue #81 

@MoniMoledo @aishwaryakapse 
As a training process, I will make some small PR for you two to review. Ask as many questions as you can 😄

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/asterix/TwitterViewsManagerActor.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/asterix/TwitterViewsManagerActor.scala
@@ -1,5 +1,7 @@
 package edu.uci.ics.cloudberry.zion.asterix
 
+import java.security.MessageDigest
+
 import akka.actor.{Actor, ActorRef, Props}
 import edu.uci.ics.cloudberry.zion.actor.{ViewMetaRecord, ViewsManagerActor}
 import edu.uci.ics.cloudberry.zion.common.Config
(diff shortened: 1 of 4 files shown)


aishwaryakapse:
Whats the best way to learn AsterixDB? I am trying to understand what the query does.

JavierJia:
@aishwaryakapse you can follow the [documentation](https://ci.apache.org/projects/asterixdb/index.html), set up a small server, write some queries, and then if you are interested in, you can also visit the [papers](http://asterixdb.ics.uci.edu/publications.html).

You can also check the `zion` module of cloudberry to see how we are using AsterixDB.

aishwaryakapse:
Thank you very much.... I will go through the documentation. It looks exciting... :)

zion/src/main/scala/edu/uci/ics/cloudberry/zion/asterix/TwitterViewsManagerActor.scala
```diff
@@ -18,7 +20,7 @@ class TwitterViewsManagerActor(val conn: AsterixConnection, override val sourceA
 
   override def getViewKey(query: DBQuery): String = {
     val keyword = query.predicates.find(_.isInstanceOf[KeywordPredicate]).map(_.asInstanceOf[KeywordPredicate].keywords.head).getOrElse("")
-    sourceName + "_" + keyword
+    sourceName + "_" + MessageDigest.getInstance("MD5").digest(keyword.getBytes("UTF-8")).map("%02x" format _).mkString
```
MoniMoledo: What is MD5 ?

JavierJia: [`MD5`](https://en.wikipedia.org/wiki/MD5) is a hash function to generate a unique checksum for a given bytes array. We are using it here to differentiate different keywords. Previously we were directly using `keyword` itself, but it won't work for the non-ASCII UTF-8 character. Because we are using this `key` to generate AQL dataset names. By using the HEX string of `MD5` value will work in AQL.

MoniMoledo:
:+1:

Issue #82, opened 2016-06-13
Handle update view failed case


JavierJia:
considering build the feed network automatically instead using the insert statement.

PR #87, opened 2016-06-16
Change insert to upsert to skip the duplicate key error in AsterixDB

JavierJia:
The `insert` statement will fail and stop the rest of insertions if it hit the duplicate key scenarios. We change it to use `upsert` to keep ingesting while ignore such failures.

It partially solved issue #82 

@MoniMoledo @aishwaryakapse


JavierJia:
I'll go ahead and merge to try a demo first.

Issue #88, opened 2016-06-16
Split the AQL queries to send them one by one


JavierJia:
Different UI component will send multiple AQL to AsterixDB. Currently, we are packing them together as one AQL file to save the round trip time. But it will complicate the code structure, moreover, the time spend on the hashtags is too long comparing to the other two. We should split the request one by one to not let all of them fail just because of one statement fails.

PR #90, opened 2016-06-19
Split a large AQL request to atomic part

JavierJia:
Summary:

We send the `byTime`, `byMap`, and `byHashtag` aggregations in one AQL. The problem is that if one of it failed, the other part would also fail. The UI won't get updates. By the PR, we split this large AQL into three AQLs in parallel. Thus, the `zion` will still respond to `neo` even some of them failed.

Related Issue:
#88 

It looks big, but most of it is about corresponding tests utility changes.
@aishwaryakapse


JavierJia:
The failure is caused by the indeterministic order. 😞 
Stay tuned.

PR #118, opened 2016-07-06
Disable the empty string query

JavierJia:
We support the empty-string query, which will draw the global views. However, the *DB will have some problem to maintain this view. Temporarily, we disable the empty string query to avoid such problem.

@starmon00


JavierJia:
@starmon00, I will go ahead and merge, yet we will still do the code review later.

Issue #105, opened 2016-06-23
Disable interpolation in time line


JavierJia:

starmon00:
@starmon00

PR #119, opened 2016-07-06
Fix barchart

starmon00:
This PR changed the linechart into a barchart in the front end webpage.

Related issues: #105 

@JavierJia

### DIFF
--- a/neo/public/javascripts/timeseries/controllers.js
+++ b/neo/public/javascripts/timeseries/controllers.js
@@ -32,7 +32,7 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
   .directive('timeSeries', function (Asterix) {
     var margin = {
       top: 10,
-      right: 10,
+      right: 30,
       bottom: 30,
       left: 50
     };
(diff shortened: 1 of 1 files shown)


neo/public/javascripts/timeseries/controllers.js
```diff
@@ -74,14 +77,15 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
               .html(minDate.getFullYear()+"-"+(minDate.getMonth()+1)+"-"+minDate.getDate());
 
             timeSeries
-              .renderArea(true)
```
JavierJia: just curious, why should we remove this line? though I'm not sure what it is doing?

JavierJia:
👍  looks good.

JavierJia:
@starmon00 I think the gap still have some problem, I pushed it online: http://cloudberry.ics.uci.edu/
You can see that the bar is too narrow yet the gap is large, I think the gap should be narrow.

starmon00:
@JavierJia ok let me add a new line of code that modifies the gap.

Issue #124, opened 2016-07-07
Add a reset button


JavierJia:
... to 
1. return to the center of the map without extra zoom out or drag.
2. reset to the entire time line.

PR #132, opened 2016-07-15
Added TimeSeries Reset Button

starmon00:
PR to add reset button for TimeSeries

@JavierJia 

#124


neo/public/javascripts/timeseries/controllers.js
```diff
@@ -70,6 +70,18 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
...
             var minDate = timeDimension.bottom(1)[0].time;
             var maxDate = timeDimension.top(1)[0].time;
+
+            chart.append('button')
+                .text("Reset")
+                .style("border", "none")
+                .style("background-color", "Transparent")
+                .style("color", "blue")
+                .style("position", "relative")
+                .style("bottom", "110px")
```
JavierJia: is it possible to use the relative size? 
sometimes we need to show the demo on a projector which has lower resolution, it is better to use `%` or something. Or you can lower your resolution to see if it looks OK.

JavierJia:
👍

PR #133, opened 2016-07-15
Reset button 

starmon00:
PR for reset button to be relative by %

@JavierJia 

#124


JavierJia:
why this is a new PR? when you push to the original remote branch it will be automatically updated. no need to create new PR. now you'd have to merge with master to fix the conflict...

JavierJia:
It will be fine to keep this one :-) otherwise, I need to revert the PR which will be worse.

PR #135, opened 2016-07-15
Update news collection

kaiyim:


JavierJia:
why are you open an new pr?

Issue #120, opened 2016-07-06
Collecting news articles related to Zika


chenlica:
We want to find web sites that provide news about Zika and get that info.

Please do some research about what site can be crawled easily.

@kaiyim

JavierJia:
@kaiyim you can follow this [post](http://www.thegeekstuff.com/2009/06/15-practical-crontab-examples/) to create a cronjob every hour to run the news collector.

PR #139, opened 2016-07-19
update news collection bash script

kaiyim:
update bash script to run in linux, and specify how to run the script hourly in cron job.
@JavierJia 
#120


JavierJia:
👍  you can sync with isg master and then merge it.

Issue #66, opened 2016-06-06
Collecting Zika-related emails from ProMED


chenlica:
http://www.promedmail.org/

Write a program to automatically retrieve Zika-related info from the service.

@JavierJia

aishwaryakapse:
@aishwaryakapse

chenlica:
@aishwaryakapse : welcome aboard!

aishwaryakapse:
Crawljax crawler written in java worked for 2 pages of "http://www.promedmail.org".
Both BFS and DFS approach implemented.
Statistics and states stored in file system.
Error: org.openqa.selenium.WebDriverException: java.lang.InterruptedException: sleep interrupted.

PR #149, opened 2016-07-29
Code for getting archiveNumbers using headers

aishwaryakapse:
Got Archive numbers in a set for page 1. Need to get number of pages to write similar queries for other pages. 

Need to incorporate this code with scrapy code to start crawling all the pages with the archive numbers in the set.

#66 

@JavierJia @chenlica @MoniMoledo @kaiyim


noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,78 @@
...
+	response = urllib2.urlopen(request)
+	return response
+
+def getQuery1Page1(start_date, end_date):
+	query1 = {"kwby1":'summary', 'search':'zika', 'show_us':'1', 'feed_id':'1'}
+	query1['date1'] = start_date
+	query1['date2'] = end_date
+	return query1
+	
+def getQuery2Page2(start_date, end_date):
```
JavierJia: all those `getQuery*page*` is exactly the same. should it be one function just called `getNext`?

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,78 @@
...
+	return listId
+
+if __name__ == '__main__':
+	
+	archNo = set([])
+	# get today's date
+	end_date = datetime.date.today()
+	start_date = datetime.datetime(2007, 01, 01)
+	end_date = end_date.strftime("%m/%d/%Y")
+	start_date = start_date.strftime("%m/%d/%Y")
```
JavierJia: too verbose. it's better to merge two assignment.
`start_date =datetime.datetime(2007, 01, 01).strftime("%m/%d/%Y")`

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,78 @@
...
+	filename = "./promedsearchresult.html.gzip"
```
JavierJia: it should be a parameter.

JavierJia: could you use [some lib](http://stackoverflow.com/questions/1543652/python-gzip-is-there-a-way-to-decompress-from-a-string) to unzip it from string other than the gzip file? it is too expensive.

noah/src/main/scala/edu/uci/ics/cloudberry/noah/crawler/promed_mail.scala
```diff
@@ -0,0 +1,8 @@
+package edu.uci.ics.cloudberry.noah.crawler
```
JavierJia: remove this file.

JavierJia:
what should we do with this PR???

aishwaryakapse:
I am extremely sorry. This code was wrong. I by-mistake deleted this branch. could you please delete this pull request?

JavierJia:
no worries. you can close this PR.

PR #150, opened 2016-08-01
Update twitter model

JavierJia:
implement the interface of actor model.

#151 

@MoniMoledo @aishwaryakapse 

It is the interface design, which will not contain many test cases. I will add more gradually.


MoniMoledo:
:+1:

PR #152, opened 2016-08-02
Final code to be run as cron job weekly

aishwaryakapse:
Saved previous two pages data by tweaking the code a bit and stored the results.

This code works for page 1.

#66 

@JavierJia @MoniMoledo @kaiyim @chenlica

### DIFF
--- a/noah/src/main/resources/crawler/promed_mail.py
+++ b/noah/src/main/resources/crawler/promed_mail.py
@@ -1,0 +1,86 @@
+import urllib2
+import urllib
+import datetime
+import zlib
+import re
+import os
+
+def getPages(query) :
+    headers = {}
+    headers['User-Agent'] = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36"
+    headers['X-Requested-With'] = 'XMLHttpRequest'
+    headers['Referer'] = "http://promedmail.org/"
+    headers['Accept'] = "application/json, text/javascript, */*"
+    headers['Accept-Encoding'] = "gzip, deflate, sdch"
+    headers['Accept-Language'] = "en-US,en;q=0.8"
+    headers['Connection'] = "keep-alive"
+    headers['Content-Type'] = "application/x-www-form-urlencoded"
+    headers['Host'] = 'promedmail.org'
+    url = "http://www.promedmail.org/ajax/runSearch.php?"
+    request = urllib2.Request(url+query, headers = headers)
+    response = urllib2.urlopen(request)
+    return response
+
+def getNext(start_date, end_date):
+    query = {'kwby1':'summary', 'search':'zika', 'show_us':'1', 'feed_id':'1'}
+    query['date1'] = start_date
+    query['date2'] = end_date
+    return query
+
+def request(filename, query):
+    with open(filename, "wb") as writeFile:
+        response = getPages(urllib.urlencode(query))
+        writeFile.write(zlib.decompress(response.read(), 16+zlib.MAX_WBITS))
+
+def getArchiveNosAndPages(fileName):
+    with open(fileName, 'rb') as f:
+        file_content = f.read()
+    listId = re.findall(r"\D(\d{8})"+"."+"(\d{4,8})\D", file_content)
+    return listId
+
+def getEachMail(archNo):
+    rheaders = {}
+    rheaders['User-Agent'] = "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36"
+    rheaders['X-Requested-With'] = 'XMLHttpRequest'
+    rheaders['Referer'] = "http://promedmail.org/"
+    rheaders['Accept'] = "application/json, text/javascript, */*"
+    rheaders['Accept-Encoding'] = "gzip, deflate, sdch"
+    rheaders['Accept-Language'] = "en-US,en;q=0.8"
+    rheaders['Connection'] = "keep-alive"
+    rheaders['Content-Type'] = "application/x-www-form-urlencoded"
+    rheaders['Host'] = 'promedmail.org'
+    url = "http://promedmail.org/ajax/getPost.php?alert_id="
+    directoryName = 'PromedResult'
+    
+    if not os.path.exists(directoryName):
+        os.makedirs(directoryName)
+        completePath = os.path.abspath(directoryName)
+    else:
+        completePath = os.path.abspath(directoryName)
+        
+    for id in archNo:
+        fname = id +'.html'
+        completeName = os.path.join(completePath, fname)
+        request = urllib2.Request(url+id, headers = rheaders)
+        response = urllib2.urlopen(request)
+
+        if not os.path.isfile(completeName):
+            with open(completeName, 'wb') as f:
+                f.write(zlib.decompress(response.read(), 16+zlib.MAX_WBITS))
+        else:
+            pass         
+
+if __name__ == '__main__':
+    archNo = set()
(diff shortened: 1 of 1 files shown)


JavierJia:
what the relation of this PR with #149 ?

aishwaryakapse:
This PR is different. The code has been edited. There are no separate functions for pages 2 and 3. As I mentioned today, I have already given the TextDB team the pages 2 and 3 of promedmail. So I wrote a general code for getting just the latest page i.e; page 1 data which we will be running as a cron job every week. Also, there is logic in this code to avoid duplicate entries. I will write a shell script to run this code once you give this a thumbs up and make it run every week as a cron job. The storage of these files is still an issue.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,86 @@
+import urllib2
+import urllib
+import datetime
+import zlib
+import re
+import os
+
+def getPages(query) :
+    headers = {}
```
JavierJia: since this `header` is static for all calls, you can just it a global variable.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,86 @@
...
+        if not os.path.isfile(completeName):
+            with open(completeName, 'wb') as f:
+                f.write(zlib.decompress(response.read(), 16+zlib.MAX_WBITS))
+        else:
+            pass         
+
+if __name__ == '__main__':
+    archNo = set()
+    end_date = datetime.date.today().strftime("%m/%d/%Y")
+    start_date = datetime.datetime(2007, 01, 01).strftime("%m/%d/%Y")
```
JavierJia: if it runs weekly, this `start_date` should be configurable as `end_date - 7days`.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,86 @@
...
+        writeFile.write(zlib.decompress(response.read(), 16+zlib.MAX_WBITS))
+
+def getArchiveNosAndPages(fileName):
+    with open(fileName, 'rb') as f:
+        file_content = f.read()
+    listId = re.findall(r"\D(\d{8})"+"."+"(\d{4,8})\D", file_content)
+    return listId
+
+def getEachMail(archNo):
+    rheaders = {}
```
JavierJia: Isn't it the same `header` as before?

aishwaryakapse: Yes. It is same.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,86 @@
...
+        completePath = os.path.abspath(directoryName)
+        
+    for id in archNo:
+        fname = id +'.html'
+        completeName = os.path.join(completePath, fname)
+        request = urllib2.Request(url+id, headers = rheaders)
+        response = urllib2.urlopen(request)
+
+        if not os.path.isfile(completeName):
+            with open(completeName, 'wb') as f:
```
JavierJia: try to reuse `request()` function?

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,86 @@
...
+    rheaders['X-Requested-With'] = 'XMLHttpRequest'
+    rheaders['Referer'] = "http://promedmail.org/"
+    rheaders['Accept'] = "application/json, text/javascript, */*"
+    rheaders['Accept-Encoding'] = "gzip, deflate, sdch"
+    rheaders['Accept-Language'] = "en-US,en;q=0.8"
+    rheaders['Connection'] = "keep-alive"
+    rheaders['Content-Type'] = "application/x-www-form-urlencoded"
+    rheaders['Host'] = 'promedmail.org'
+    url = "http://promedmail.org/ajax/getPost.php?alert_id="
+    directoryName = 'PromedResult'
```
JavierJia: it should be parameter that configurable by cmd.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -5,50 +5,32 @@
...
-    rheaders['Accept'] = "application/json, text/javascript, */*"
-    rheaders['Accept-Encoding'] = "gzip, deflate, sdch"
-    rheaders['Accept-Language'] = "en-US,en;q=0.8"
-    rheaders['Connection'] = "keep-alive"
-    rheaders['Content-Type'] = "application/x-www-form-urlencoded"
-    rheaders['Host'] = 'promedmail.org'
+def getEachMail(archNo, headers, directoryName):
+    
     url = "http://promedmail.org/ajax/getPost.php?alert_id="
     directoryName = 'PromedResult'
```
JavierJia: why you change the var that passed in?

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -72,15 +54,40 @@ def getEachMail(archNo):
...
+    headers['Connection'] = "keep-alive"
+    headers['Content-Type'] = "application/x-www-form-urlencoded"
+    headers['Host'] = 'promedmail.org'
+    
+    end_date = datetime.date.today()
+    start_date = end_date - datetime.timedelta(7)
+    end_date = end_date.strftime("%m/%d/%Y")
+    start_date = start_date.strftime("%m/%d/%Y")
+    filename = "promedsearchresult.html"
+    dirName = "PromedResult"
```
JavierJia: can you make all these vars passing from cmd? 
it could be something like
`python promed_mail.py --start 20160101 --end 20160107 --dir local` ?

aishwaryakapse: we can send the dir name from cmd line... I don't think it is needed to pass end date and start date as cmd parameters because the code is taking end date as todays date( the date this program will be run) and start date is calculated by the program as 7 days before todays date.

I didn't reuse the request function because I am overwriting the page file for every new page. This I am doing because our program will run every week and if we do not overwrite the previous weeks page, there will be too many unnecessary files. I cannot overwrite the email files as I need to make sure all files are there and also there are no duplicates.

JavierJia: The reason taking the start and end date into the cmd parameter is that the `7 days` period is not fixed. what if the zika is bursting and we want to collect it daily? the bash can be easily changed without touching the source code.

noah/src/main/resources/crawler/promed_mail.py
```diff
@@ -0,0 +1,93 @@
...
+def getIds(fileName):
+    with open(fileName, 'rb') as f:
+        file_content = f.read()
+    listId = re.findall(r"\D(\d{8})"+"."+"(\d{4,8})\D", file_content)
+    return listId
+
+def getEachMail(archNo, headers, directoryName):
+    
+    url = "http://promedmail.org/ajax/getPost.php?alert_id="
+    directoryName = 'PromedResult'
```
JavierJia: why changing the given `directoryName`?

noah/src/main/resources/crawler/promed_mail.py
aishwaryakapse: That can be changed by changing the cron job time. the start date can be constant, i.e; the date we run the program. So we can make the time interval to be entered via cmd. such as 7 days or 2 days. That can be done. But the thing is, we are entering the website from the search tab which is not modified in real time. It gets modified after some time ( maybe days after the actual update). So, I feel that keeping the start date interval a constant value and just changing the end date by changing the cron time sounds a more user-friendly way. Even if we run it every day, it will show the results of 7 days from that date. And since we have already taken care of avoiding duplicates, it shouldn't matter. 


If we want the cmd parameters, I can add them as one. Its just a line of code.

JavierJia: I agree with your point, instead of passing the start and end time, we can pass the `interval` parameter. Good idea!

PR #159, opened 2016-08-08
Removed the absolute path and allowed for local file storage

aishwaryakapse:
Removed the code that was determining the absolute path of the current directory. 

Allowed for storage in user specified directory.

@JavierJia @kaiyim @MoniMoledo @chenlica


JavierJia:
👍

Issue #151, opened 2016-08-01
Use new JSON data model to implement the Cloudberry


JavierJia:
After implementing the data model in #30 , we need to push the logic to the replace the current ad-hoc middleware. 

including
- [x] updating the actor model
- [x] updating the front-end query model
- [x] updating the front-end parsing logic #59

PR #161, opened 2016-08-09
Add create view query

JavierJia:
to build the view dataset, we need to calculate the view schema and the corresponding create aql parser. 

#151 

@MoniMoledo


MoniMoledo:
:+1:

PR #163, opened 2016-08-09
Added Shell script to run promed_mail crawler and set up cron job

aishwaryakapse:
Added cron job

@JavierJia @chenlica @kaiyim @MoniMoledo

### DIFF
--- a/noah/src/main/resources/crawler/runPromed.sh
+++ b/noah/src/main/resources/crawler/runPromed.sh
@@ -1,3 +1,8 @@
-#Run the scrapy crawler for promedmail.org
+#!/bin/bash
 
-scrapy crawl promedSpider
+#Run promed_mail.py
+
+cd /home/akapse/cloudberry/noah/src/main/resources/crawler
+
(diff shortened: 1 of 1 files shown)


noah/src/main/resources/crawler/runPromed.sh
```diff
@@ -1,3 +1,8 @@
-#Run the scrapy crawler for promedmail.org
+#!/bin/bash
+
+#Run promed_mail.py
+
+cd /home/akapse/cloudberry/noah/src/main/resources/crawler
```
JavierJia: we should not put the absolute link to the code base.

noah/src/main/resources/crawler/runPromed.sh
```diff
@@ -1,3 +1,8 @@
...
+
+python promed_mail.py /home/akapse/PromedResult 1
```
JavierJia: ditto

JavierJia:
I think we don't need to submit the actual script that we are using on kiwi machine.

PR #164, opened 2016-08-10
Aql parser update - Aggregate function without group by

MoniMoledo:
Update AQLParser to support Group Statements without the by statement  in order to extend the use of aggregate functions.

@JavierJia

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLFuncVisitor.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLFuncVisitor.scala
@@ -205,26 +205,37 @@ object AQLFuncVisitor {
     */
   def translateAggrFunc(field: Field,
                         func: AggregateFunc,
-                        aqlExpr: String
+                        aqlExpr: String,
+                        isAggrOnly: Boolean
                        ): (DataType.DataType, String, String, String) = {
-    val newvar = s"${aqlExpr.split('.')(0)}aggr";
+    val parameter = StringBuilder.newBuilder
+    val wrap = StringBuilder.newBuilder
+    if (isAggrOnly) {
+      parameter.append("")
+      wrap.append("")
+    } else {
+      val newvar = s"${aqlExpr.split('.')(0)}aggr";
+      parameter.append(s"$newvar")
+      wrap.append(")")
+    }
+
     func match {
       case Count =>
         if (field.dataType != DataType.Record) throw new QueryParsingException("count requires to aggregate on the record bag")
-        (DataType.Number, s"count($newvar)", newvar, s"let $newvar := $aqlExpr")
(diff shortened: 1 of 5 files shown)


zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLFuncVisitor.scala
```diff
@@ -205,26 +205,37 @@ object AQLFuncVisitor {
...
-    val newvar = s"${aqlExpr.split('.')(0)}aggr";
+    val parameter = StringBuilder.newBuilder
+    val wrap = StringBuilder.newBuilder
+    if (isAggrOnly) {
+      parameter.append("")
+      wrap.append("")
+    } else {
+      val newvar = s"${aqlExpr.split('.')(0)}aggr";
+      parameter.append(s"$newvar")
+      wrap.append(")")
```
JavierJia: since the `wrap` always contains `)`, can you remove the `)` and write the return string from
`s"count($parameter$wrap"`  to `s"count($parameter$wrap)" ? otherwise, i will be crazy to find the paired `)`

JavierJia: I think I'm wrong, never mind.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLQueryParserTest.scala
```diff
@@ -399,33 +399,66 @@ class AQLQueryParserTest extends Specification {
...
+          |return $t.'id')""".stripMargin)
+    }
+
+    "translate a count cardinality query with filter without group by" in {
+      val filter = Seq(timeFilter)
+      val group = GroupStatement(Seq.empty, Seq(aggrCount))
+      val query = new Query(dataset = TwitterDataSet, filter = filter, groups = Some(group))
+      val result = parser.parse(query, schema)
+      removeEmptyLine(result) must_== unifyNewLine(
+        """count(
```
JavierJia: could you also try `min` ?

JavierJia:
I still feel that the code flow doesn't very clear. We'd better reshape the original code part in which case we might need to introduce the totally new branch at very beginning rather than inside the function. 
let's discuss it tomorrow.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLQueryParserTest.scala
MoniMoledo: Sure! Done

MoniMoledo:
I feel the same. I've created a different branch to deal with it in AQLFunctionVisitor and AQLQueryParser. However, still left some code on "parseSelect" to handle the case there is a select sttmt inside an aggregate function, that stills doesn't feel right.

PR #167, opened 2016-08-11
Add GlobalAggregateStatement to the query

MoniMoledo:
Add GlobalAggregateStatement to the query in order to support max, min, count.. functions

@JavierJia

PR #168, opened 2016-08-12
Add tests for DataManager

JavierJia:
Add actor model related tests. 
The missing part is to get the global stats of one dataset which is rely on #167 . 

@MoniMoledo


MoniMoledo:
:+1:

PR #172, opened 2016-08-17
Restructure cloudberry + timeseries bug fix

starmon00:
PR for restructuring cloudberry by creating divs for different parts of cloudberry and fixing timeseries zoom bug

@JavierJia


neo/app/views/index.scala.html
```diff
@@ -1,15 +1,31 @@
...
+      <a href="https://www.youtube.com/watch?v=XwlRm0jcsU4" target="_blank" class="btn btn-info">How to use this demo</a>
+    </div>
+
+  </div>
+
+  <div class="SearchBar">
+    <search-bar></search-bar>
+  </div>
+
+  <div class="TimeSeries">
```
JavierJia: let's unify the js class name pattern. I think it's more common to use all-lower-case with `-` in js.

neo/public/javascripts/searchbar/controllers.js
```diff
@@ -16,6 +16,7 @@ angular.module('cloudberry.util', ['cloudberry.common'])
       restrict: "E",
       controller: 'SearchCtrl',
       template: [
+        '<style> .btn.btn-info{position : relative; left:105%; z-index: -3;} </style>',
```
JavierJia: remove this line.

neo/app/views/index.scala.html
```diff
@@ -1,15 +1,31 @@
 @(title: String) @main(title){
 <div xmlns="http://www.w3.org/1999/html" ng-controller="AppCtrl">
 
-  <map lat="39.5" lng="-96.35" zoom="4"></map>
+  <div class="Map">
```
JavierJia: I think `<map>` should have its own id? then this div may not need?

JavierJia: may be `map-group` to be clear?

neo/app/views/index.scala.html
```diff
@@ -1,15 +1,31 @@
...
+      <img src ="https://asterixdb.apache.org/img/asterixdb_tm.png" id="logo_img">
+    </div>
+
+    <div id='how-to'>
+      <a href="https://www.youtube.com/watch?v=XwlRm0jcsU4" target="_blank" class="btn btn-info">How to use this demo</a>
+    </div>
+
+  </div>
+
+  <div class="SearchBar">
```
JavierJia: search-bar should have its own `id`, this `div` may not be required?

neo/app/views/index.scala.html
```diff
@@ -1,15 +1,31 @@
...
-  <map lat="39.5" lng="-96.35" zoom="4"></map>
+  <div class="map-group">
 
-  <search-bar></search-bar>
 
-  <time-series id="chart"></time-series>
+    <map lat="39.5" lng="-96.35" zoom="4"></map>
 
+    <div id="logo">
+      <img src ="https://asterixdb.apache.org/img/asterixdb_tm.png" id="logo_img">
```
JavierJia: let's remove the logo for now. we haven't decided which one to use. it's fine to keep the `div` here.

neo/app/views/index.scala.html
```diff
@@ -1,15 +1,29 @@
...
+      <a href="https://www.youtube.com/watch?v=XwlRm0jcsU4" target="_blank" class="btn btn-info">How to use this demo</a>
+    </div>
+
+  </div>
 
   <search-bar></search-bar>
 
-  <time-series id="chart"></time-series>
 
+  <div class="Time-Series">
```
JavierJia: lower case?

neo/public/stylesheets/main.css
```diff
@@ -97,3 +99,19 @@ search-bar {
...
+  z-index = 0;
+  position: absolute;
+  top:2%;
+  left:10%;
+  width : 15%;
+  height: 10%;
+
+}
+
+.btn.btn-info{
```
JavierJia: why do we need this setting?

starmon00: I left it in there so when they add the logo it will be auto positioned to the spot. I can remove it if we don't need it.

JavierJia: got it. but to make it explicit, could you rename the class to `logo` than the general `btn`?

Issue #53, opened 2016-05-30
Zoom in to city level details


JavierJia:
The frontend can ask the shapes as needed, not like the state and county shapes which were loaded at the beginning.

JavierJia:
@zonghengma

JavierJia:
@zonghenma, you can take a look at the js code inside  `neo/public/javascripts/map/controllers.js`

zonghengma:
If just load the entire city data, the webpage gets stuck for around 10 seconds on my machine when it is first loaded. It also gets stuck for 5 seconds or so when the map is zoomed into city level.

JavierJia:
sounds cool. could send a PR (pull request) about the change? we don't need to merge, but we can have a check about the code.

zonghengma:
Great. I'll do it as soon as I get home.

JavierJia:
No rush 😄

chenlica:
Cool.  It verified the conjecture that loading too much city data to the browser will slow down the response.  For our record, what's the size of the city data?

After @JavierJia checks your code, please start working on the idea we discussed last time: the frontend loads data from the middleware those cities in a certain region.  We also need to implement a function in the middleware to find cities in a region by scanning all the cities.

zonghengma:
Updates: I have attempted to load the entire city data in the beginning. However, this makes the webpage to get stuck completely for about 10 sec in the beginning as well as when zoom into city levels. Once in city level, the map becomes very unresponsive. We can conclude that this is not an acceptable way. I am currently trying to only load the city data inside within the map currently being shown in the window. This contains modifications in the middleware as well. I'll post new updates here as soon as I have new findings.

zonghengma:
Updates:
The middleware can now calculates the mid-points of the cities on first launch. 
Front-end can send a get request whenever the map is dragged in city level.
Todo:
Store the mid-points someway in order and return to front-end as need.

@JavierJia Do we need to change the calculation of mid-points of states and counties from the front-end to middleware? It is kind of messy if we calculate some mid-points in the middleware and some in the front-end.

chenlica:
Cool!  Looking forward to the demo with the city-level details!

JavierJia:
@zonghengma Processing by the front-end is an ideal case which can save a lot msg passing to middleware. Use middleware is a fallback because the city is too big to process in front-end.

JavierJia:
@zonghengma can upload a PR for review?

zonghengma:
[...]

PR #206, opened 2016-10-09
Attempted to load the entire city data in the beginning.

zonghengma:
Sorry for the delay. There is a problem that the map stops responding after zoom into city level. I was trying to debug it but I have not found a solution. In addition, the cities cannot fill the entire map. There are a lot empty spaces. I think it may because there are unincorporated areas. It looks kind of ugly.


JavierJia:
I think 1) we don't have the entire city json, 2) the city concept only valid within the real city, the rural part won't be covered by the city shape. Thus, the city level can never cover the entire map like the county level does.

neo/public/javascripts/map/controllers.js
```diff
@@ -98,7 +105,20 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
-          if ($scope.status.zoomLevel > 5) {
+          if($scope.status.zoomLevel > 7) {
+            $scope.status.logicLevel = 'city';
+            if (!$scope.status.init) {
+              resetGeoIds($scope.bounds, $scope.geojsonData.city, 'cityID');
+              Asterix.parameters.geoLevel = 'city';
+              Asterix.queryType = 'zoom';
+              Asterix.query(Asterix.parameters, Asterix.queryType);
+            }
+            if ($scope.polygons.statePolygons) {
```
JavierJia: since we have three levels now, we should also check the `|| $scope.polygons.countyPolygons` ?

neo/public/javascripts/map/controllers.js
```diff
@@ -107,6 +127,7 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
               Asterix.query(Asterix.parameters, Asterix.queryType);
             }
             if($scope.polygons.statePolygons) {
+              $scope.map.removeLayer($scope.polygons.cityPolygons);
```
JavierJia: same here, the if condition should check the other two levels.

neo/public/javascripts/map/controllers.js
```diff
@@ -119,6 +140,7 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
               Asterix.query(Asterix.parameters, Asterix.queryType);
             }
             if($scope.polygons.countyPolygons) {
+              $scope.map.removeLayer($scope.polygons.cityPolygons);
```
JavierJia: ditto.

neo/public/javascripts/map/controllers.js
```diff
@@ -129,7 +151,15 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
       $scope.$on("leafletDirectiveMap.dragend", function() {
         if (!$scope.status.init) {
           $scope.bounds = $scope.map.getBounds();
-          var geoData = ($scope.status.logicLevel === 'state') ? $scope.geojsonData.state : $scope.geojsonData.county;
+          var geoData;
+          if($scope.status.logicLevel === 'state'){
+            geoData = $scope.geojsonData.state;
+          } else if($scope.status.logicLevel === 'county'){
+            geoData = $scope.geojsonData.county;
+          }else if($scope.status.logicLevel === 'city'){
```
JavierJia: there should be a space between `}` and `else`.

neo/public/javascripts/map/controllers.js
```diff
@@ -129,7 +151,15 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
+            geoData = $scope.geojsonData.city;
+          }
```
JavierJia: we should have an `else` branch, log the error `logicLevel` and set to the default one. (probably the `state` one)

neo/public/javascripts/map/controllers.js
```diff
@@ -129,7 +151,15 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
+          // var geoData = ($scope.status.logicLevel === 'state') ? $scope.geojsonData.state : $scope.geojsonData.county;
```
JavierJia: remove the unused lines.

neo/public/javascripts/map/controllers.js
```diff
@@ -323,18 +364,33 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
         // draw
         $scope.polygons.statePolygons.setStyle(style);
       } else if ($scope.status.logicLevel == "county" && $scope.geojsonData.county) {
-        angular.forEach($scope.geojsonData.county.features, function(d) {
+          angular.forEach($scope.geojsonData.county.features, function(d) {
+            if (d.properties.count)
```
JavierJia: the code in `county` and `city` (and probably the `state`) levels are quite similar. could you find a way to combine them?

zonghengma:
@JavierJia Thanks for the corrections. Did you find any problems that can cause the map to stop responding? Currently, when I zoom into city level, the map stop responding to drag actions. It takes no actions to queries as well. It gets back to normal when I zoom out to county/state levels. I got a error in the browser console but I don't know what does it mean.

zonghengma:
filehelper_1476073153320_11

JavierJia:
That's error msg comes from the `timeserial` component which doesn't look like relevant. 
We can discuss this kind of problem on slack.

chenlica:
Slack is a better place for this type of discussion. Or a F2F discussion :-)

zonghengma:
I have fixed the issues in the previous pull request.

JavierJia:
If you address a comment, you should reply `done` to that comment. Then reviewers can easily track the new changes.

neo/public/javascripts/map/controllers.js
zonghengma: Done.

neo/public/javascripts/map/controllers.js
zonghengma: Done.

neo/public/javascripts/map/controllers.js
zonghengma: Done

neo/public/javascripts/map/controllers.js
zonghengma: Done

neo/public/javascripts/map/controllers.js
zonghengma: Done

neo/public/javascripts/map/controllers.js
zonghengma: Done

neo/public/javascripts/map/controllers.js
zonghengma: Added a FIXME comment here. I'll try to combine them later as I know more about JS

neo/public/javascripts/map/controllers.js
```diff
@@ -120,16 +145,30 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
-          var geoData = ($scope.status.logicLevel === 'state') ? $scope.geojsonData.state : $scope.geojsonData.county;
+          var geoData;
+          if ($scope.status.logicLevel === 'state') {
+            geoData = $scope.geojsonData.state;
+          } else if ($scope.status.logicLevel === 'county') {
+            geoData = $scope.geojsonData.county;
+          } else if ($scope.status.logicLevel === 'city') {
+            geoData = $scope.geojsonData.city;
+          } else {
+            console.log("Error: Illegal value of logicLevel, set to default: state")
```
JavierJia: If that is an error, you can use `console.error` function.

JavierJia:
This PR showed the problem of loading the entire city into the front-end. It also serves the purpose of excising the code review process. We don't need to merge it though. 

You can check out a new branch from the master and start implementing the on-demand loading logic we discussed before. I'm closing this PR.

JavierJia:
#53

PR #213, opened 2016-10-25
Calculate the centroid of cities in middleware when it starts

zonghengma:
It is what I have at this point. I will update it soon.

Things left:
Store the mid-points someway in order after calculation;
Binary search the mid-points to return to front-end.

### DIFF
--- a/geotag.sh
+++ b/geotag.sh
@@ -19,7 +19,9 @@
 
 set -o nounset                              # Treat unset variables as an error
 
+thread=$1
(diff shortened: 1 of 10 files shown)


JavierJia:
Sounds good. But for each PR please make it as atomic as possible. We don't need to finish one task by just one PR. Usually it has multiple steps, and each step should be separate PR.

For this PR specifically, please just submit the middleware changes.

JavierJia:
and again, usually even in your local repo, you should create a feature branch instead of using the `master`

zonghengma:
Closed due to additional file changes in the PR that may cause confusion. See PR #214

Issue #94, opened 2016-06-21
Create the namespace for each running Middleware


JavierJia:
... so that each developer won't be impacted by the other since we are using the same test cluster.
Otherwise, we should use the docker container locally to solve this issue, which would be another good approach.

JavierJia:
Solved by use dockerize AsterixDB.

PR #216, opened 2016-10-31
Simplify the demo set up by using AsterixDB docker

JavierJia:
This PR utilize the AsterixDB docker image to create a local cluster, then feed the sampled twitter data into it, which makes the demo much easier to set up than before.


JavierJia:
#94

PR #220, opened 2016-11-12
Frontend centroid calculation method

zonghengma:
While developing the code for centroid calculation for the middleware, we found a minor error in the method we used for centroid calculation in the front-end. This PR fixed the error.


waans11:
Can you show us an example of the given error?

zonghengma:
@waans11 As you can see, the old code calculate the controids by averaging the X and Y coordinates for all vertices of the polygons. The correct way is to find the min and max value for X and Y coordinates, and X coordinate of the centroid is (minX + maxX) / 2, same thing for Y coordinate. 

The wrong code actually worked since the centroid calculated by the old method is really close to the real centroid. It was not a very big issue but I used the correct method in the middleware, I changed it here for consistency.

PR #225, opened 2016-11-22
Do binary search in middleware while searching for cities within a boundary.

zonghengma:
The middle-ware can do binary search. I'll compare the times of running binary search and scaning all the cities. Related to issue #228


chenlica:
@zonghengma The PR shows "Some checks were not successful."  Please fix it.

@waans11 Can you review this PR per our discussion?

waans11:
@zonghengma : can you put some comments in the code? I'm having hard time to understand the logic, variable name, etc. General question to all: I wonder "no-commnet" is our policy on Cloudberry?

chenlica:
"No comment" is NEVER our policy.  I am a believer of providing succinct, high-quality comments.

@zonghengma Please do as suggested.

zonghengma:
@chenlica Yes I have located the error and fixed it. I'll update the PR with some other improvements.

zonghengma:
@waans11 Adding them soon

zonghengma:
A quick update:
Now I use just one binary search and a for loop (instead of filter), but this implementation is even slower than when we scan the entire list without binary search. I believe something in the loop make this implementation slow. I'm currently investigating the cause and hopefully there could be a fast solution that only use one binary search.

I commit both the implementations to the PR. findCityNew() is the slow one that use one binary search and loop, findCity() is the faster one that use two binary searches and filter.

chenlica:
@waans11 Can you take a look?

@zonghengma Please notice that the check tailed.

waans11:
@chenlica: I will try after everything is fixed.

zonghengma:
@chenlica Yeah I saw the check failed. It says it's better to use switch case instead of if - else in my code, but actually there's no reason to use switch case in that place.

waans11:
@zonghengma: I checked the code briefly. One naive thought: can we combine the binary search (for starting point) + filter? I am not familiar with scala at all. So, I guess maybe filter method does play a role here. So, does filter need to be applied to the whole array? or can we specify a starting index?

zonghengma:
@waans11 The filter need to be applied on an array. If you have the starting index and ending index, you can first slice the array and the apply the filter on the sliced array. 

However, if you only have the starting index, you can only cut out the front part from the array but not the "tail" part. Filter cannot be "break" in the middle like a loop.

So for the new implementation, I used one binary search to get the starting index, and uses a for loop. I think the for loop and the if-else statements are the reason for it to be slow.

waans11:
@zonghengma: So, we have three possible cases. 
1) scan and filter (the current way)
2) find head (binary search) and apply if-else until we see the tail (original suggested).
3) find head (binary search) and tail (binary search).

I think the method 3) would be faster if the cost of a single binary search is less than the scan and filtering the region (if-else) cost. Since we had 30,000 cities, O(log_2 30,000) = 14. We need to have 14 searches for a single binary search case.

My question to you is: on average, how many elements that we need to proceed? That is, what's the difference between head and tail (index difference)? I think it's more than 14.

zonghengma:
@waans11 You are right. (3) is much faster than (1) and (2). I don't know the exact number but it is definitely much more than 14.

The findCity method in this PR implements method (3), and findCityNew method implements method (2).

For (2) and (3), we both need to filter(or scan) the list for latitude since the list is sorted only based on longitude. The difference is in (2) I used a for loop and in (3) I used the filter method. The only thing I don't understand now is how exactly the filter method works and why (2) is much slower than (3) and even (1). It should do comparisons for each element in the list similar to a for-loop, but (2) is much slower than (3) (I remembered is about 8ms vs 70ms).

@JavierJia do you think we should keep (3) for now? For (3) when we filter the list, we only need to compare the latitude, but for (2) we need to compare both latitude and longitude. It seems that even if the cost of a for-loop and the filter method is the same, (3) still saves some comparisons.

chenlica:
@zonghengma I thought (2) and (3) should have similar performance.  If (3) is much faster, it may be because of the additional comparison for each element in (2).  Let's choose the fastest one!

waans11:
@zonghengma: For exact comparison between filter() and our if-else check, I think what you can try is two things.
(3-1). find head (binary search) and tail (binary search). use filter() to filter out the cities that are not within the region (latitude).
(3-2). find head (binary search) and tail (binary search). use if-else() to filter out the cities that are not within the region (latitude).   

Also, for the exact numbers, I think you can print out the starting index and the ending index when the middle-ware actually receives the request from the front-end. I think the number will be greater than 14 since we have more than 14 cities in the map region that we see in the screen.

JavierJia:
👍  @chenlica 's suggestion.

zonghengma:
Thanks! I'll do the test and update the PR

zonghengma:
I have modified this PR to only contain code for (3-1)

@waans11 the running time for (3-1) and (3-2) is 84ms vs 7ms. I commented out the list.builder part of the code and the time did not reduce so much, so it is the loop and if-else make the code slow. That number for the highest level in city-level is typically around 1000, so its far more than 14.

chenlica:
If @waans11 and @JavierJia both approve it, we can do the merge.

If "codacy" still complains, maybe we can get rid of this plugin?  @JavierJia

waans11:
@zonghengma: _"the running time for (3-1) and (3-2) is 84ms vs 7ms."_ for this, what you mean is that 3-1.filter() method takes 7ms while 3-2.if-else() method takes 84ms? That's why you want to keep 3-1? Right?

zonghengma:
@waans11 Yes. I did this comparison between if-else with a loop(3-2) and filter(3-1). (3-1) is much faster so I kept (3-1)

JavierJia:
@chenlica Codacy check is not mandatory. It won't block us from merging. And the committer should know that it often gives good suggestions, it's better to take a look.

neo/app/controllers/Application.scala
```diff
@@ -151,12 +153,40 @@ object Application{
   }
 
   def findCity(neLat: Double, swLat: Double, neLng: Double, swLng: Double, cities: List[JsValue]) =  {
-    //TODO: Do binary search
-    val citiesWithinBoundary = cities.filter{
-      city =>
-        (city \ CentroidLatitude).as[Double] <= neLat && (city \ CentroidLatitude).as[Double] >= swLat.toDouble && (city \ CentroidLongitude).as[Double] <= neLng.toDouble && (city \ CentroidLongitude).as[Double] >= swLng.toDouble
+    /*
```
JavierJia: We should write Javadoc that the comments are on top of the function.
```
/**
  *  // IntelliJ will automatically fill in the necessary values.
  */ 
def findCity(neLat: Double, swLat: Double, neLng: Double, swLng: Double, cities: List[JsValue])
```

JavierJia: Every function should explicitly declare the return type.
`def findCity(neLat: Double, swLat: Double, neLng: Double, swLng: Double, cities: List[JsValue])` ==> 
`def findCity(neLat: Double, swLat: Double, neLng: Double, swLng: Double, cities: List[JsValue]) : JsArray`

neo/app/controllers/Application.scala
```diff
@@ -151,12 +153,40 @@ object Application{
...
+      val citiesWithinBoundary = cities.slice(startIndex, endIndex).filter {
+          city =>
+            (city \ CentroidLatitude).as[Double] <= neLat && (city \ CentroidLatitude).as[Double] >= swLat.toDouble
+      }
+      val response = header + (Features -> Json.toJson(citiesWithinBoundary))
+      Json.toJson(response)
+    }
+  }
+
+  def binarySearch(cities: List[JsValue], start: Int, end: Int, target: Double) : Int = {
```
JavierJia: `targetLat`? or `targetLog`?

neo/app/controllers/Application.scala
```diff
@@ -151,12 +153,40 @@ object Application{
...
+    if (start == end) {
+      start
+    } else {
+      val thisIndex = (start + end) / 2
+      val thisCity = cities.apply(thisIndex)
```
JavierJia: `cities.apply(thisIndex)` can be written as `cities(thisIndex)`

neo/app/controllers/Application.scala
```diff
@@ -151,12 +153,40 @@ object Application{
...
+      val centroidLongitude = (thisCity \ CentroidLongitude).as[Double]
+      if (centroidLongitude > target){
+        binarySearch(cities, start, thisIndex, target)
+      }
```
JavierJia: usually, we take the following convention.

```
if () {
} else if {
} else {
}
```

zonghengma:
I have added the Javadoc fixed the minor issues

neo/app/controllers/Application.scala
```diff
@@ -150,13 +150,53 @@ object Application{
...
+    * @param swLat Latitude of the SouthWest point of the boundary
+    * @param neLng Latitude of the NorthEast point of the boundary
+    * @param swLng Latitude of the SouthWest point of the boundary
+    * @param cities List of all cities
+    * @return List of cities which centroids is in current boundary
+    */
+  def findCity(neLat: Double, swLat: Double, neLng: Double, swLng: Double, cities: List[JsValue]) : JsValue =  {
+    /*
+      Use binary search twice to find two breakpoints (head and tail) to take out all cities whose longitude are in the range,
+      then scan those cities one by one for latitude.
```
JavierJia: these comments should go to Javadoc between line 153 and 154

JavierJia:
👍

waans11:
👍

PR #242, opened 2017-01-30
Add eclipse dev support, and fix kafka version

luochen01:


project/plugins.sbt
```diff
@@ -15,5 +15,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-coffeescript" % "1.0.0")
 
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.3.5")
 
+addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "5.1.0")
```
JavierJia: It won't be any file generated if I didn't run `sbt eclipse`, right?

JavierJia:
I saw you are working on your master which is often deprecated. It's better to work on the feature branches even in your own repo.

project/plugins.sbt
luochen01: yes, run sbt eclipse will automatically download the plugin and generate eclipse project files.

Issue #194, opened 2016-09-26
Implement the Lookup statement


JavierJia:
There are some missing features in the AQLGenerator. The most important one is the `lookup` statement. 

Here is one example 

```
    { "dataset":"twitter",
      "lookup": [{
          "joinKey":"geo.city",
          "dataset":"US_population",
          "lookupKey":"city",
          "select":"population",
          "as" : "us_population"
},
         {
          "joinKey":"geo.city",
          "dataset":"tweet_user",
          "lookupKey":"city",
          "select":"population"}
],
      "filter":[
        { "field":"create_at", 
          "relation":"inRange",    
          "values":["11-01-2016","11-08-2016"]},
        { "field":"text",
          "relation":"contains",
          "values":["zika","virus"]}],
      "group":{
        "by":[
           {"field":"geo.state", "as" : "state"],
        "aggregate":[
           { "field": "*", "apply": "count"}
           { "field": "population", "apply": "sum", 
             "as":"population"}]}}]
  }
```
In this example, we lookup the `US_population` dataset and join two dataset with the condition of `twitter.geo.city = US_population.city` and only append the `population` field to the current dataset (`twitter`)

Finally, it will group the records to different states. Within each group, it can apply the `count` for `*` field as usual. Plus, it can also apply the `sum` function on the newly appended `population` field.

JavierJia:
We use the Play JSON library to parse the JSON to the Query object. You can find the corresponding document [here](https://www.playframework.com/documentation/2.5.x/ScalaJson#Json)

JavierJia:
- [ ] Translate the lookup JSON to lookup Scala query object. You can start from the `JSONParser` and its test. 
- [ ] Generate the Twitter user data and test the actual performance of lookup

nishadg:
Link to wiki : https://github.com/ISG-ICS/cloudberry/wiki/Lookup-Functionality

PR #244, opened 2017-02-02
[Issue #194] [WIP] Implement parsing of lookup queries

nishadg:
Issue #194 
AQLGenerator will now be able to generate AQL query for single and multiple lookup statements. A lookup statement is used to join two datasets using left outer join.
Signature of generate(), parseQuery() and parseLookup() has been changed. They now accept a map of (dataset name -> schema) instead of only schema. 

Need feedback on

1.  Function signature change.
2. Test cases written in AQLGeneratorTest for this functionality.


JavierJia:
Looks great at the first glance. 
Since it's really difficult to verify the generated statement, you need to make sure that you've created corresponding AsterixDB dataset and those generated AQL can return correct result.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,70 @@ class AQLGeneratorTest extends Specification {
...
+      val selectStatement = SelectStatement(Seq.empty, 0, 0, selectValues)
+      val lookup = LookupStatement(Seq("geo_tag.stateID"), populationDataSet, Seq("stateID"), selectValues,
+        selectValues)
+      val filter = Seq(textFilter)
+      val query = new Query(TwitterDataSet, Seq(lookup), filter, Seq.empty, select = Some(selectStatement))
+      val result = parser.generate(query, Map(TwitterDataSet -> schema, populationDataSet -> populationSchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |for $l0 in dataset twitter.US_population
```
JavierJia: @waans11 could you verify if the join order is correct? Usually, the `lookup` table is very small and should be the build part. I forget which should go first.

JavierJia: in this example, the `us_population` is the lookup table.

chenlica:
@waans11 I hope this is a good exercise for you to get familiar with the codebase....

chenlica:
@nishadg and @JavierJia I believe we need to decide and follow a naming convention for PRs.  I suggest: (1) If there is an issue, use [Issue ###] as the prefix; (2) Make the subject more informative; and (3) make the description more detailed.

nishadg:
Noted.

JavierJia:
About the first one, as long as the description mentions the issue number, GitHub will automatically link it. 
Totally agree on the other two. 🍕

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -121,9 +135,9 @@ class AQLGenerator extends IQLGenerator {
         """.stripMargin
       )
       //TODO check if the vars are duplicated
```
JavierJia: is this `TODO` addressed?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -121,9 +135,9 @@ class AQLGenerator extends IQLGenerator {
...
-      val field: Field = ??? // get field from lookup table
+      val fieldMap: Map[String, Field] = schemaMap.flatMap(_._2.fieldMap) // get fields from lookup and query table
```
JavierJia: if we have two tables with the same field name, what will be the flatted result?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -121,9 +135,9 @@ class AQLGenerator extends IQLGenerator {
...
       producedVar ++= lookup.as.zip(lookup.selectValues).map { p =>
-        p._1 -> AQLVar(new Field(p._1, field.dataType), s"$lookupVar.${p._2}")
+        p._1 -> AQLVar(new Field(p._1, fieldMap(p._2).dataType), s"$lookupVar.${p._2}")
```
JavierJia: using `p` as a pair is always hard to understand what P actually is. I suggest to use `case match` as following

```
... map { case ( reName: String, oriName: String) =>
   reName -> ***** oriName ***
}
```
Then we will know what is `_1` and what is `_2`.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,70 @@ class AQLGeneratorTest extends Specification {
...
+          |where $l0.stateID = $t.'geo_tag'.'stateID'
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |limit 0
+          |offset 0
+          |return
+          |{ '*': $t, 'population': $l0.population}
+        """.stripMargin.trim
+      )
+    }
```
JavierJia: Can you add one more test that using `groupby` sentence to group by the state and get the `sum` of the generated population field?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
nishadg: It will have only one field. there will not be duplicates.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
nishadg: Okay. Will do.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
JavierJia: then there is the problem, right? we removed one or more field(s) just because it has the same name with the first one?

nishadg: Ok. This is a bit tricky. Will think about this problem.

nishadg: @JavierJia How will the parser know which schema the select name is referring to?
eg. Emp(id, name, salary, dept_id) & Dept(id, name)
the lookup JSON will be
{
          "joinKey":"dept_id",
          "dataset":"Dept",
          "lookupKey":"id",
          "select":"name",
          "as" : "name"
}

Which name should we return here?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
waans11: the first one : outer (probe)
the second one : inner (build)

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -21,6 +21,19 @@ class AQLGenerator extends IQLGenerator {
...
   }
 
+  def generate(query: IQuery, schemaMap: Map[String,Schema]): String = {
+    query match {
+      case q: Query =>
+        validateQuery(q)
+        parseQuery(q, schemaMap)
+      case q: CreateView => parseCreate(q, schemaMap(query.dataset))
+      case q: AppendView => parseAppend(q, schemaMap(query.dataset))
+      case q: UpsertRecord => parseUpsert(q, schemaMap(query.dataset))
```
waans11: upsertRecord -> upsertView? Not sure whether what this does.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
waans11: Maybe we need to add one more field? "lookupDataset"?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
waans11: Put the us_population in the first for.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,70 @@ class AQLGeneratorTest extends Specification {
...
+      val query = new Query(TwitterDataSet, Seq(lookup), filter, Seq.empty, select = Some(selectStatement))
+      val result = parser.generate(query, Map(TwitterDataSet -> schema, populationDataSet -> populationSchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |for $l0 in dataset twitter.US_population
+          |where $l0.stateID = $t.'geo_tag'.'stateID'
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |limit 0
```
waans11: What does 'limit 0' do?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
nishadg: But we can have select values from both source and lookup dataset.
{
"joinKey":"dept_id",
"dataset":"Dept",
"lookupKey":"id",
"select":"name, salary",
"as" : "name, salary"
}
Please correct me if I'm wrong.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/LiteracyDataStore.scala
```diff
@@ -0,0 +1,21 @@
+package edu.uci.ics.cloudberry.zion.model.impl
+
+import edu.uci.ics.cloudberry.zion.model.schema.{NumberField, Schema, TimeField}
+
+/**
+  * Created by nishadgurav on 2/1/17.
```
waans11: It's better to say some description here about this object.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/PopulationDataStore.scala
```diff
@@ -0,0 +1,21 @@
+package edu.uci.ics.cloudberry.zion.model.impl
+
+import edu.uci.ics.cloudberry.zion.model.schema._
+
+/**
+  * Created by nishadgurav on 1/19/17.
```
waans11: Also, here, it would be really nice if we put some description.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
nishadg: Since we cannot have a select without limit, I added limit 0. This means query will not return anything.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
JavierJia: source fields only selected in the separate `select` statement. the lookup statement only select the columns from lookup table.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
nishadg: Need more clarification regarding this todo. In what case will the vars be duplicated?

JavierJia: I think it is being addressed now. we can remove this todo. 👍

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/PopulationDataStore.scala
```diff
@@ -0,0 +1,24 @@
...
```
JavierJia: my general suggestion is to remove the `created by author` template. since we are using git, that is easy to track.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
nishadg: The parse lookup will generate the for loop for lookup statements with the where clause. This is then appended to the source dataset for.
So if we put the lookup fors first it will look like this:
`for $l0 in dataset twitter.US_population
where $l0.stateID = $t.'geo_tag'.'stateID'
for $t in dataset twitter.ds_tweet`

JavierJia: definitely not correct. but we have to swap the order, otherwise the performance will be really really bad. 
I think we need to split the `for` part and the `where` part, and return both strings. then put the `for $t in dataset twitter.ds_tweet` in the middle. 

Otherwise, we need to change AsterixDB 😄

JavierJia: wait, I think we are in the wrong direction. Here we want a *left-out-join* not an equal join.
I found one using SQL++ [here](https://ci.apache.org/projects/asterixdb/sqlpp/manual.html#Left_outer_joins), @waans11 do you know what the corresponding AQL syntax for that ?

waans11: for $user in dataset MugshotUsers
where $user.user-since >= datetime('2010-07-22T00:00:00')
  and $user.user-since <= datetime('2012-07-29T23:59:59')
return {
  "uname" : $user.name,
  "messages" :
    for $message in dataset MugshotMessages
    where $message.author-id = $user.id
    return $message.message
};

waans11: For our case:
          |for $t in dataset twitter.ds_tweet
          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
          |and contains($t.'text', "virus")
          |limit 0
          |offset 0
          |return
          |{ '*': $t, 'population': 
          for $l0 in dataset twitter.US_population
          where $t.'geo_tag'.'stateID' /* +indexnl */ = $l0.stateID
          return $l0.population}

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
...
+      val group = GroupStatement(Seq(byState), Seq(AggregateStatement("population", Sum, "sum")))
+      val lookup = LookupStatement(Seq("geo_tag.stateID"), populationDataSet, Seq("stateId"), selectValues,
+        selectValues)
+      val filter = Seq(textFilter)
+      val query = new Query(TwitterDataSet, Seq(lookup), filter, Seq.empty, groups = Some(group))
+      val result = parser.generate(query, Map(TwitterDataSet -> schema, populationDataSet -> populationSchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |for $l0 in dataset twitter.US_population
```
JavierJia: As a baseline implementation for the lookup is that we compose an augemented dataset by append new columes, and then operate the rest of the statements (e.g., select, groupby)

```
for $t in dataset twitter.ds_tweetwhere
where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
and contains($t.'text', "virus")
return {
'creat_at' : $t.create_at
'id' : $t.id
....  /// **flatten all $t's field by reading them from the schema.** 
'population' : ( for $l0 in dataset twitter.US_population 
                         where $l0.stateID = $t.'geo_tag'.'stateID'
                         return $l0.population) [0]  /// **here since it's lookup, let assume**
                                                                     ///  **if it gonna hit multiple values,** 
                                                                     /// **they should be same, then we just pick one.**
'other-lookups' : (xxxx)[0]
}

/// **then we can use the same logic as if we are group on a single table.** 
let $l0aggr := $l0.population
group by $g0 := $t.geo_tag.stateID with $l0aggr
return {
  'state' : $g0,'sum' : sum($l0aggr)
}
```
Above example is not guaranteed to be optimal. But let's first make a working one 🥇

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -102,31 +115,36 @@ class AQLGenerator extends IQLGenerator {
   }
 
   private def parseLookup(lookups: Seq[LookupStatement],
-                          varMap: Map[String, AQLVar]
+                          varMap: Map[String, AQLVar],
+                          schemaMap: Map[String, Schema]
                          ): (String, Map[String, AQLVar]) = {
-    val sb = StringBuilder.newBuilder
+//    val sb = StringBuilder.newBuilder
```
JavierJia: remove not used line?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -102,31 +115,36 @@ class AQLGenerator extends IQLGenerator {
...
-      //TODO check if the vars are duplicated
-      val field: Field = ??? // get field from lookup table
-      producedVar ++= lookup.as.zip(lookup.selectValues).map { p =>
-        p._1 -> AQLVar(new Field(p._1, field.dataType), s"$lookupVar.${p._2}")
-      }
+           |return $lookupVar.${lookup.selectValues.head})[0]
+        """.stripMargin.trim
+
+
+      val fieldMap: Map[String, Field] = schemaMap(lookup.dataset).fieldMap // get fields from lookup dataset
```
JavierJia: rename the `fieldMap` to `lookupTableFieldMap` to remove the comment  ?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
           |}""".stripMargin.trim)
     }
 
+    "translate lookup one table with one join key" in {
+      val populationDataSet = PopulationDataStore.DatasetName
+      val populationSchema = PopulationDataStore.PopulationSchema
+
+      val selectStatement = SelectStatement(Seq.empty, 0, 0, Seq("*", "population"))
+      val lookup = LookupStatement(Seq("geo_tag.stateID"), populationDataSet, Seq("stateId"), Seq("population"),
+        Seq("population"))
```
JavierJia: for the multiple parameters case, 
1. usually we want a better alignment. 
2. we can explicitly write the parameter's name to make the unamed varaibles meaningful. e.g., you can write
```
val lookup = LookupStatement(
sourceKeys = Seq("geo_tag.stateID"), 
dataset = populationDataSet, 
lookupKeys = Seq("stateId"), 
selectValues= Seq("population"),
as = Seq("population"))
```

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
...
+      val result = parser.generate(query, Map(TwitterDataSet -> schema, populationDataSet -> populationSchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |limit 0
+          |offset 0
+          |return
+          |{ '*': $t, 'population': (for $l0 in dataset twitter.US_population
```
JavierJia: I think we need to flatten this `*`, otherwise, the returned nested result is very difficult for user to handle.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
...
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |limit 0
+          |offset 0
+          |return
+          |{ '*': $t, 'twitter.US_population': for $l0 in dataset twitter.US_population
+          |where $t.'geo_tag'.'stateID' /* +indexnl */ = $l0.stateId
```
JavierJia: here we have another `*/`, maybe try another way of commenting?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
...
+        literacyDataSet -> literacySchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |limit 0
+          |offset 0
+          |return
+          |{ '*': $t, 'population': (for $l0 in dataset twitter.US_population
```
JavierJia: same here, we should flatten the `*`

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,123 @@ class AQLGeneratorTest extends Specification {
...
+        selectValues)
+      val filter = Seq(textFilter)
+      val query = new Query(TwitterDataSet, Seq(lookup), filter, Seq.empty, groups = Some(group))
+      val result = parser.generate(query, Map(TwitterDataSet -> schema, populationDataSet -> populationSchema))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |for $t in dataset twitter.ds_tweet
+          |where similarity-jaccard(word-tokens($t.'text'), word-tokens('zika')) > 0.0
+          |and contains($t.'text', "virus")
+          |let $population_aggr := (for $l0 in dataset twitter.US_population
```
JavierJia: cool!

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -21,6 +21,19 @@ class AQLGenerator extends IQLGenerator {
     }
   }
 
+  def generate(query: IQuery, schemaMap: Map[String, Schema]): String = {
```
waans11: Can you put a comment for this function? What does it do?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
waans11: Answer for this?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -102,31 +116,35 @@ class AQLGenerator extends IQLGenerator {
...
+           |(for $lookupVar in dataset ${lookup.dataset}
            |where ${keyZip.mkString(" and ")}
-        """.stripMargin
-      )
-      //TODO check if the vars are duplicated
-      val field: Field = ??? // get field from lookup table
-      producedVar ++= lookup.as.zip(lookup.selectValues).map { p =>
-        p._1 -> AQLVar(new Field(p._1, field.dataType), s"$lookupVar.${p._2}")
-      }
+           |return $lookupVar.${lookup.selectValues.head})[0]
```
waans11: Can we put a comment for the reason why we only use the first element? Maybe as todos?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGeneratorTest.scala
```diff
@@ -530,6 +530,143 @@ class AQLGeneratorTest extends Specification {
...
+          |limit 0
+          |offset 0
+          |return
+          |{ 'text': $t.'text', 'geo_tag.countyID': $t.'geo_tag'.'countyID', 'favorite_count': $t.'favorite_count', 'in_reply_to_user': $t.'in_reply_to_user', 'geo_tag.cityID': $t.'geo_tag'.'cityID', 'coordinate': $t.'coordinate', 'is_retweet': $t.'is_retweet', 'user.id': $t.'user'.'id', 'lang': $t.'lang', 'population': (for $l0 in dataset twitter.US_population
+          |where $t.'geo_tag'.'stateID' /* +indexnl */ = $l0.stateId
+          |return $l0.population)[0], 'in_reply_to_status': $t.'in_reply_to_status', 'user_mentions': $t.'user_mentions', 'create_at': $t.'create_at', 'user.status_count': $t.'user'.'status_count', 'geo_tag.stateID': $t.'geo_tag'.'stateID', 'retweet_count': $t.'retweet_count', 'id': $t.'id', 'geo': $t, 'hashtags': $t.'hashtags'}
+        """.stripMargin.trim
+      )
+    }
+
```
waans11: Let's remove commented lines. If we want to keep it, then please leave the reason.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/PopulationDataStore.scala
waans11: Agreed. Please remove all "author" tag.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
nishadg: I am not sure about this. @JavierJia can you help us out here?

JavierJia: Oh, it's supposed to be used when we need to update the view's metadata. we can make it clear when @haochen07  clean up the dataset register work.

JavierJia:
@nishadg  it's better to sperate the JSON parser logic to a separate PR. 
especially when I've already approved the previous changes.

nishadg:
@JavierJia Sorry for the mix up. I have done a reset head and pushed

JavierJia:
@nishadg  can you check the Travis test [failure](https://travis-ci.org/ISG-ICS/cloudberry/builds/204322224)?
It says 

> java.lang.ClassCastException: edu.uci.ics.cloudberry.zion.model.schema.Query cannot be cast to edu.uci.ics.cloudberry.zion.model.schema.UpsertRecord (DataStoreManagerTest.scala:106)

@waans11 it will be nice you  can give a quick review again so that we can merge this one and not block @nishadg's work 😁

JavierJia:
Looks good. please go ahead and merge 👍

Issue #230, opened 2017-01-07
Replace the actor message passing in TweetMap demo by using "/berry" http request


JavierJia:
Currently, in the front-end server, the `NeoActor` is sending the message directly to the `berryClient`. It works because the front-end and the middleware live in the same server which will not be the case if we want to make Cloudberry a general purpose middleware. 

Now, since we have the `/berry` API, the `NeoActor` should send an HTTP request to `http://hostname:port/berry` instead of an actor message and then receive a stream of HTTP chunked results.

You can read this [documentation](https://www.playframework.com/documentation/2.5.x/ScalaWS#Processing-large-responses) about how to use it to handle a stream of responses.

JavierJia:
@haochen07

JavierJia:
You can start from the non-streaming fashion first. :-)

haochen07:
This Issue is done. Do I need to close it ? @JavierJia

JavierJia:
please do 📪

haochen07:
which means close it ?

JavierJia:
haha, yes, please feel free to close it.

PR #248, opened 2017-02-08
[Issue #230] Switch communication channel between NeoActor and Berry from actor messages to Http request/response

haochen07:
This PR is related to Issue #230. 

**Purpose**
Now the front-end NeoActor talks to middleware through direct actor messages. This is not the general case because front-end and middleware are actually in the same server. Instead of using actor messages, this update enables NeoActor to send http request to `/Berry` API and get streamed http response. 

**Implementations**

- Implement http post request and streaming response to exchange information between NeoActor and BerryClient, instead of actor messages.
- Use akka streaming model to implement streaming response.
- Update global configurations to define maximum length of frame can be transport in each data chunk. (default value 8 MB)
- Remove NeoActorTransformer, which is no longer useful.
- Enable the logging of any possible errors in http connection and stream response reception.
- Update test case for NeoActor and the sbt build file for importing issue.


neo/conf/application.conf
```diff
@@ -76,6 +76,7 @@ bounded-mailbox {
 }
 
 neo.timeout.initial = "10 minutes"
+neo.stream.max.frame.length = "8388608 Bytes"
```
JavierJia: why this number?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
```diff
@@ -31,5 +33,10 @@ object Config {
     FiniteDuration(split(0).toLong, split(1))
   }
 
+  def parseMemoryPair(memoryString: String): Int = {
```
JavierJia: `parseMemoryConfiguration` to be precise?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
```diff
@@ -12,6 +12,8 @@ class Config(config: Configuration) {
 
   val AwaitInitial = config.getString("neo.timeout.initial").map(parseTimePair).getOrElse(10 minutes)
 
+  val maxFrameLength = config.getString("neo.stream.max.frame.length").map(parseMemoryPair).getOrElse(8388608)
```
JavierJia: use Capital letter for global static values, though codacy don't like it :-)

JavierJia: rename to `MaxFrameLengthForNeoWS` to be more precise? 
the global static values deserves a explicit good name

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,107 @@
...
           "batch" -> JsArray(groupTimePlaceTags),
           "option" -> JsObject(Seq("sliceMillis" -> JsNumber(2000))
-          )))
-        berryClient.tell(BerryClient.Request(json, NeoTransformer("batch")), out)
+        )))
+
+        // Dealing with response on timing, geo and hashtags: slicing
+        val batchResponse: Future[StreamedResponse] =
+          ws.url(url).withMethod("POST").withBody(batchJson).stream()
+        batchResponse.map{ response =>
```
JavierJia: it looks like the two response processing codes are very similar, can you make one function call that can be used in both cases by passing different parameters?

neo/conf/application.conf
haochen07: This is for maximum 8MB per chunk. I am not very sure of the number. It cannot be too small to hold a single piece of data returned by Berry, nor can be so large that will take memory and time to transport. Maybe we need experiments to find the optimal value. In short, I need suggestions on that.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
haochen07: I changed this in the new commit

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
haochen07: I changed the function name in the new commit

neo/app/actor/NeoActor.scala
haochen07: I reconstruct the code for this in the new commit

neo/conf/application.conf
JavierJia: I think let's just stick with this number for now. 
in terms of "good code", it's better to enable something like
`neo.stream.max.frame.length = "8MB"`
or simply
`neo.stream.max.frame.megabytes = 8`
to make more sense 😁

haochen07: I fixed the code.
For the test case of Neo, can we compare the json result received in NeoActor to the json result sent by middleware, to test if we get all data right?

JavierJia:
Regarding the test case, you can take a look of the DataSetAgentTest.scala, which uses a mock connection to return a predefined value and test if the out actor can receive expected msg. 🤞

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
   import NeoActor._
-
   implicit val timeout: Timeout = Timeout(20.minutes)
 
   override def receive: Receive = {
     case json: JsValue =>
       json.validate[UserRequest].map { userRequest =>
-        val allRequest = generateCBerryRequest(userRequest)
+
+        // Generate json in berry format
```
JavierJia: code tells. if your code needs a lot of this kind of comments, that means the code is not expressive enough.

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
         import RequestType._
+        val berryRequest = generateCBerryRequest(userRequest)
+
+        // Get Berry url address
```
JavierJia: ditto

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+        val url = "http://" + host + "/berry"
 
-        // Example of sending the non-slicing query using ask pattern
-        val sampleJson = allRequest(Sample)
-        val fAnswer = (berryClient ? sampleJson).mapTo[JsValue].map(json => Json.obj("key" -> "sample", "value" -> json))
-        fAnswer pipeTo out
+        // Dealing with response on samples: slicing
+        handleHttpTransport(url, "sample", berryRequest(Sample))
```
JavierJia: in order to remove the above comment, you can write a function
`handleSamplingResponse`
later you can have another function `handleSlicingResponse`. Then people will know what you are doing.

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
   }
-}
-
-object NeoActor {
-  def props(out: ActorRef, berryClientProp: Props)(implicit ec: ExecutionContext) = Props(new NeoActor(out, berryClientProp))
 
-  case class NeoTransformer(key: String) extends BerryClient.IPostTransform {
-    override def transform(jsValue: JsValue): JsValue = {
-      Json.obj("key" -> key, "value" -> jsValue)
+  private def handleHttpTransport(url: String, requestType: String, requestBody: JsValue): Unit = {
```
JavierJia: `requestType: String` should be a defined enum  type instead of string.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
```diff
@@ -31,5 +33,10 @@ object Config {
     FiniteDuration(split(0).toLong, split(1))
   }
 
+  def parseFrameLengthLimit(memoryString: String): Int = {
```
JavierJia: then what if people are writing `8 G`? or `8M` (without space?)

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+    // Post request and get future response
+    val response : Future[StreamedResponse] =
+      ws.url(url).withMethod("POST").withBody(requestBody).stream()
+
+    // Wrap streaming response and send to webpage
+    response.map{ res =>
+      if(res.headers.status == 200){
+        try {
+          // Define sink of data stream
```
JavierJia: code tells.

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+          val sink = Sink.foreach[ByteString] { bytes =>
+            val json = Json.parse(bytes.utf8String)
+            out ! Json.obj("key" -> requestType, "value" -> json)
+          }
+          // streaming data pipeline
+          res.body.via(Framing.delimiter(ByteString("\n"), maximumFrameLength = config.MaxFrameLengthForNeoWS, allowTruncation = true))
+                  .runWith(sink)
+                  .onFailure { case e => Logger.logger.error(e.toString) }
+        }
+        catch {
```
JavierJia: we usually merge this line with `}`
`} catch {`

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+          case e: Throwable => Logger.logger.error("stream receiving error: " + e.getMessage)
+        }
+      }
+      // Gateway error
+      else {
```
JavierJia: merge with `}` to 
`} else {`

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
-        val json = JsObject(Seq(
+        //  Dealing with response on timing, geo and hashtags: slicing
+        val groupTimePlaceTags = Seq(ByTime, ByPlace, ByHashTag).map(berryRequest(_))
+        val batchJson = JsObject(Seq(
           "batch" -> JsArray(groupTimePlaceTags),
           "option" -> JsObject(Seq("sliceMillis" -> JsNumber(2000))
-          )))
-        berryClient.tell(BerryClient.Request(json, NeoTransformer("batch")), out)
+        )))
+        handleHttpTransport(url, "batch", batchJson)
```
JavierJia: should we create two constant `val` for `batch` and `sample` ?

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+    // Wrap streaming response and send to webpage
+    response.map{ res =>
+      if(res.headers.status == 200){
+        try {
+          // Define sink of data stream
+          val sink = Sink.foreach[ByteString] { bytes =>
+            val json = Json.parse(bytes.utf8String)
+            out ! Json.obj("key" -> requestType, "value" -> json)
+          }
+          // streaming data pipeline
```
JavierJia: code tells.

neo/app/actor/NeoActor.scala
```diff
@@ -1,57 +1,88 @@
...
+          res.body.via(Framing.delimiter(ByteString("\n"), maximumFrameLength = config.MaxFrameLengthForNeoWS, allowTruncation = true))
+                  .runWith(sink)
+                  .onFailure { case e => Logger.logger.error(e.toString) }
+        }
+        catch {
+          case e: Throwable => Logger.logger.error("stream receiving error: " + e.getMessage)
```
JavierJia: `"stream receiving error:` ==> `"NeoActor websocket post receiving ....`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
```diff
@@ -31,5 +34,36 @@ object Config {
     FiniteDuration(split(0).toLong, split(1))
   }
 
+  object MemorySize extends Enumeration {
```
JavierJia: this object should be out of the scope of `Config` or in the companion object of `Config`.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/common/Config.scala
```diff
@@ -31,5 +34,36 @@ object Config {
...
+    val KB: Int = 1024
+    val MB: Int = 1024*1024
+    val GB: Int = 1024*1024*1024
+  }
+
+  def parseFrameLengthLimit(memoryString: String): Int = {
+    val size = memoryString.substring(memoryString.length-2)
```
JavierJia: this function should have a test.

haochen07: I tested it by changing the following value in `.conf` file: `"8MB"` `"8 MB"` `"8 TB"` `"8 GB"` `"2GB"` `"1KB"`, the results are expected. How should I present the test for this case?

JavierJia: there should be the test codes reflecting what you said above. 
[here](https://github.com/ISG-ICS/cloudberry/blob/65872b8fc21468cf612e2be1323c6a8d922f2894/zion/src/test/scala/edu/uci/ics/cloudberry/zion/common/ConfigTest.scala) is one example.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/common/ConfigTest.scala
```diff
@@ -11,6 +11,17 @@ class ConfigTest extends Specification {
       Config.parseTimePair("43 minute") must_==(43 minutes)
       Config.parseTimePair("84 minutes") must_==(84 minutes)
     }
-
+    "parseFrameLengthLimit" in {
+      Config.parseFrameLengthLimit("20 B") must_==(8 * MemorySize.MB)
+      Config.parseFrameLengthLimit("0 KB") must_==(8 * MemorySize.MB)
+      Config.parseFrameLengthLimit("0.9 KB") must_==(8 * MemorySize.MB)
+      Config.parseFrameLengthLimit("1 KB") must_==(1 * MemorySize.KB)
```
JavierJia: that's a weird setting, right? 
why I get `1K` when we give `1K`, but `8M` when we give `0.5KB`?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/common/ConfigTest.scala
```diff
@@ -11,6 +11,17 @@ class ConfigTest extends Specification {
...
+      Config.parseFrameLengthLimit("16 MB") must_==(16 * MemorySize.MB)
+      Config.parseFrameLengthLimit("16MB") must_==(16 * MemorySize.MB)
+      Config.parseFrameLengthLimit("1.5 GB") must_==(1.5 * MemorySize.GB)
+      Config.parseFrameLengthLimit("2 GB") must_==(Int.MaxValue)
+      Config.parseFrameLengthLimit("3 GB") must_==(8 * MemorySize.MB)
+      Config.parseFrameLengthLimit("5 TB") must_==(8 * MemorySize.MB)
```
JavierJia: one principle about test case is that it should be meaningful and trackable when it fails. 

so in this case, you can structure them like following
```
"parseFrameLengthLimit should give 1KB if the given value is <= 1KB" in {
 ...
}
"parseFrameLengthLimit should parse the MB and GB if the given value is valid" in {
 ...
}
"parseFrameLengthLimit should use 2G as maxmumal value" in {
 ... 
}
```

neo/app/actor/NeoActor.scala
```diff
@@ -1,63 +1,86 @@
...
+    handleResponse(url, Batch, batchJson)
+  }
+
+  private def handleResponse(url: String, requestType: RequestType.Value, requestBody: JsValue): Unit = {
+    val response : Future[StreamedResponse] =
+      ws.url(url).withMethod("POST").withBody(requestBody).stream()
+
+    response.map{ res =>
+      if(res.headers.status == 200){
+        try {
```
JavierJia: [Codacy](https://www.codacy.com/app/ISG-Cloudberry/cloudberry/pullRequest?prid=510426) complains about the `try catch` problem. 
Can you take a look?

haochen07: At my first glance, it suggests us:
1. to use `Util.Try` instead of `try`
2. our `catch` function catches all errors, which is not a good way because we should let the fatal errors continue raising.
I'll follow up those points

JavierJia: since it's also new to me. I encourage to dig a little more and we can learn some rule and summarize it on one of the wiki page.

haochen07: Based on my research, `Try` works with case class `Success` and `Failure`, 

`Try{
  someError()
} match {
  case Success(v) =>
  case Failure(e) =>
}`

is a wrapper of
 
`try{
  someError()
} catch {
  case NonFatal(e) =>
}`

The idea is that it catches all **NonFatal** exceptions and let the fatal errors continue to propagate. The advantage of using `Try` is to pipeline errors along the way. See example here [scala Try](http://www.scala-lang.org/api/2.12.x/scala/util/Try.html). 

However, in our case, one of the error we concern is if the capacity `config.MaxFrameLengthForNeoWS` is big enough to hold every single chunk of response data. This error is reported by neither of them, but caught by `.onFailure()` function at the end of streaming pipeline. And removing that function will not let `Try` or `try` catching the error. I might think that this `.onFailure()` function is sufficient in our case.

neo/test/actor/NeoActorTest.scala
```diff
@@ -99,5 +115,53 @@ class NeoActorTest extends Specification {
...
+      val jsonRequest = Json.parse(
+        s"""
+           |{
+           |  "dataset"     : "twitter.ds_tweet",
+           |  "keywords"    : ["zika"],
+           |  "timeInterval": ${JsObject(Seq(("start", JsNumber(interval.getStartMillis)), ("end", JsNumber(interval.getEndMillis))))},
+           |  "timeBin"     : "day",
+           |  "geoLevel"    : "state",
+           |  "geoIds"      : [37, 51],
+           |  "mergeResult" : ${JsBoolean(false)}
```
JavierJia: just curious, why we need a `JsBoolean` instead of direct `false` value?

neo/test/actor/NeoActorTest.scala
```diff
@@ -99,5 +115,53 @@ class NeoActorTest extends Specification {
...
+      when(ws.url(any)).thenReturn(wsRequest)
+      when(wsRequest.withMethod("POST")).thenReturn(wsRequest)
+      when(wsRequest.withBody[JsValue](any[JsValue])(any[Writeable[JsValue]])).thenReturn(wsRequest)
+      when(wsRequest.stream()).thenReturn(Future(StreamedResponse(header, source)))
+
+      val frontEnd = new TestProbe(system)
+      val neo = system.actorOf(NeoActor.props(frontEnd.ref, ws, url, Config.Default))
+      frontEnd.send(neo, jsonRequest)
+
+      val jsonExpected1a = Json.obj("key" -> Sample, "value" -> json1)
```
JavierJia: can you rename these two variables? like `jsonExpected1a` --> `ExpectedSampleJson1` ?

neo/test/actor/NeoActorTest.scala
```diff
@@ -99,5 +115,53 @@ class NeoActorTest extends Specification {
...
+      val jsonExpected1b = Json.obj("key" -> Batch, "value" -> json1)
+      val jsonExpected2a = Json.obj("key" -> Sample, "value" -> json2)
+      val jsonExpected2b = Json.obj("key" -> Batch, "value" -> json2)
```
JavierJia: ditto

neo/test/actor/NeoActorTest.scala
```diff
@@ -99,5 +115,53 @@ class NeoActorTest extends Specification {
...
+
+      val result1 = frontEnd.receiveN(2)
```
JavierJia: same thing for the result1, it's hard to pair with the expected variable.

neo/test/actor/NeoActorTest.scala
```diff
@@ -99,5 +115,53 @@ class NeoActorTest extends Specification {
...
+      result1.contains(jsonExpected1a) must_== true
+      result1.contains(jsonExpected1b) must_== true
+      val result2 = frontEnd.receiveN(2)
```
JavierJia: ditto

neo/test/actor/NeoActorTest.scala
haochen07: Because a `JsBoolean` is required in this field, I put a `JsBoolean` directly. `false` also works in this case. I change it in new commit for simplicity.

neo/test/actor/NeoActorTest.scala
haochen07: updated

neo/test/actor/NeoActorTest.scala
haochen07: updated as `ExpectedStreamingResult1`

neo/test/actor/NeoActorTest.scala
JavierJia: good to know 😄

chenlica:
Congrats! @haochen07 and @JavierJia

Issue #235, opened 2017-01-20
Show a total count of tweets in the front-end


JavierJia:
We need to show the live count of the total number of tweets somewhere in the front-end, like this [demo](https://www.mapd.com/demos/tweetmap/)

The following steps will be suggested:
- [x] Show a static count on the front-end
- [x] Make a request to the middleware, and get the total number of tweets continuously. 
- [x] In the middleware, make a call to DB daily sometime to get the actual count and also keep the update rate to emulate the changes.

PR #257, opened 2017-02-24
[issue #235] [WIP] Show the Count of the Tweets

HotLemonJuice:
Related issue: #235 
Purpose:

- Show the count of all the tweets that are displayed in the current level(state, county, or city).
- Show the total count of tweets sent by the middleware.
- Avoid overlapping between the count div and the time series div.

Implement Details:

timeseries/controller.js
- Get the sum of all the tweets in the current view from the time results
- Watch the Asterix.totalCount to change the total count displayed on the page
- Wrap the "countDiv" and the "timeSeries" with a "stats" div and use float: left on the countDiv
- Set position of the reset button in the time series to absolute so that it can adjust its position correctly according to relatively positioned parent "time-series".
- Use angular filter to format the numebr.

screen shot 2017-03-02 at 21 10 48
screen shot 2017-03-02 at 21 11 05
screen shot 2017-03-02 at 21 11 15

# Screen Recordings
count11 mov
count2 mov
count23 mov

# New styling
screen shot 2017-03-03 at 10 33 27


neo/public/javascripts/map/controllers.js
```diff
@@ -183,6 +184,39 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
       };
       $scope.controls.custom.push(info);
 
+      // TODO: change it to the real count sent from the middleware
```
JavierJia: I think this part is for the totalcount feature?

JavierJia:
Could you remove all this `totalcount` related codes? This PR should only about show the `count(  the current query's result)`. Thanks!

neo/public/javascripts/map/controllers.js
HotLemonJuice: yes, you are right. I put a fake number for now just to demo the UI. :)

JavierJia: then you can just keep that piece of logic locally. otherwise, the reviewers have to spend time on the irrelevant stuff.

JavierJia:
@HotLemonJuice any progress on this PR?

HotLemonJuice:
I have merged with the master and delete the fake number for the total count. Now it displays only the current count with black font color.

JavierJia:
The current totalcount is a real one. please understand the current changes and show both number.

neo/public/javascripts/map/controllers.js
```diff
@@ -1,5 +1,5 @@
 angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
-  .controller('MapCtrl', function($scope, $window, $http, $compile, Asterix, leafletData) {
+  .controller('MapCtrl', function($scope, $window, $http, $compile, Asterix, leafletData, $timeout) {
     $scope.result = {};
     $scope.totalCount = 0;
```
JavierJia: here is the totalCount. you can take a look at where and how it showed.

HotLemonJuice: I see. I have integrate the changes.

HotLemonJuice:
Now it shows the real total count as well. I want to avoid the count div overlap with the time series div, but they are under different parent divs (count div is under map-group, and time series is under time-series). Which way would you suggest me to move them around? Move count div into time series or move time series into map group or maybe just make a new parent div to include both?

JavierJia:
Both have pros and cons, I don't think it's difficult to try both ways. you can have a try and tell us which one is better?

neo/public/javascripts/map/controllers.js
```diff
@@ -1,7 +1,8 @@
...
+    $scope.currentTweetsCount = 0;
```
chenlica: `currentTweetsCount` -> `currentTweetCount`?

chenlica:
To make this PR easier to understand, please add a screenshot of the new UI after the change (can be part of the window).  @HotLemonJuice

neo/public/javascripts/map/controllers.js
HotLemonJuice: okay

HotLemonJuice:
@JavierJia Just a heads up. I notice the total count number will revert back to the initial value when it reaches 40 million or so.  Do you have the same issue on your machine?

JavierJia:
No. you should debug your own code to see what cause your problem. 
based on the code I saw, I think you may need to understand how the totalcount is shown in the master. then add your changes on top it.

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -1,29 +1,57 @@
...
+    // add information about the count of tweets
+    var countDiv = document.createElement("div");
+    countDiv.id = "countDiv";
+    countDiv.title = "Display the count information of Tweets";
+    let itemName = "tweets";
+    countDiv.innerHTML = [
+      '<p class="big-text"> {{ currentTweetCountString }} </p>',
+      '<span>',
+        '<span class="small-text">of</span>',
+        '<span class="big-text">&nbsp;{{ totalCountString }}</span>',
```
JavierJia: you can use the angula filter function `totalCount | number` to avoid formatting.

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -1,29 +1,57 @@
...
+        $scope.currentTweetCountString = $scope.formatNumber($scope.currentTweetCount);
 
       }
       return result_array;
     };
+
+    // add information about the count of tweets
+    var countDiv = document.createElement("div");
+    countDiv.id = "countDiv";
+    countDiv.title = "Display the count information of Tweets";
```
JavierJia: what does this title do?

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -1,29 +1,57 @@
...
+    let itemName = "tweets";
```
JavierJia: is this `itemName` variable necessary? could we just put the string there in html?

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -39,6 +67,19 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
...
     );
+
+    $scope.$watch(
+      function () {
+        return Asterix.totalCount;
+      },
+
+      function (newCount) {
+        if(newCount) {
+          $scope.totalCountString = $scope.formatNumber(newCount);
```
JavierJia: if use angular filter, there is no need to format.

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -102,9 +143,9 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
                 .text('Reset')
                 .attr('href',"#")
                 .on("click", function() { timeSeries.filterAll(); dc.redrawAll(); requestFunc(minDate, maxDate);})
-                .style("position", "inherit")
+                .style("position", "absolute")
```
JavierJia: just curious, what's the different from `absolute` and the `inherit`?

neo/public/stylesheets/main.css
```diff
@@ -31,11 +34,10 @@ body {
...
 time-series {
-  position: absolute;
-  top: 82%;
-  left: 15%;
-  width: 80%;
-  height: 15%;
+  z-index: 1000;
+  pointer-events: none;
+  display: inline-block;
+  position: relative;
```
JavierJia: I'm not familiar with css, could you teach me what's the meaning of each values? thanks

neo/public/javascripts/map/controllers.js
```diff
@@ -1,7 +1,6 @@
 angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
-  .controller('MapCtrl', function($scope, $window, $http, $compile, Asterix, leafletData) {
+  .controller('MapCtrl', function($scope, $window, $http, $compile, Asterix, leafletData, $timeout) {
```
JavierJia: remove `$timeout`?

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -1,29 +1,57 @@
...
+  .controller('TimeSeriesCtrl', function ($scope, $window, $compile, Asterix) {
     $scope.ndx = null;
     $scope.result = {};
     $scope.resultArray = [];
     $scope.d3 = $window.d3;
     $scope.dc = $window.dc;
     $scope.crossfilter = $window.crossfilter;
     $scope.empty = [];
+    $scope.currentTweetCount = 0;
+    $scope.formatNumber = function (number) {
```
JavierJia: we don't need this function.

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -1,29 +1,57 @@
...
+      return number.toLocaleString('en-US');
+    };
+    $scope.totalCountString = "0";
+    $scope.currentTweetCountString = "0";
```
JavierJia: we should only need the `int`. the string can be generated.

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: The value of the title attribute will be displayed as a tooltip when you mouse over the paragraph

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: Absolute: The element is positioned relative to its first positioned (not static) ancestor element. 
Inherit: Inherits the position property from its parent element.
So in this case, make the "a" tag's position absolute will make its positioning properties like "left", "bottom" be relative to the time-series tag whose position is set to relative. If I don't do this, the reset button on the time series will be in some strange place.

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: yes, absolutely.

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: Okay, I got it.

neo/public/javascripts/timeseries/controllers.js
JavierJia: Got it. Thanks!

neo/public/javascripts/timeseries/controllers.js
JavierJia: Got it.

JavierJia: Good to know the technique, but do you think it's necessary?

neo/public/stylesheets/main.css
```diff
@@ -31,11 +34,10 @@ body {
 }
 
 time-series {
-  position: absolute;
-  top: 82%;
-  left: 15%;
-  width: 80%;
-  height: 15%;
+  z-index: 1000;
+  pointer-events: none;
```
JavierJia: what is `pointer-event`?

JavierJia:
Looks good! And the screenshot is pretty convincing!
I just a comment about the newly added CSS code.

chenlica:
Wow, I don't expect you can even make an animated gif!  Looks great.  A minor comment: the space between the result count and the total tweet count is a bit too wide.  Can you reduce it?  @HotLemonJuice Thanks!

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: Not really, since the number is self-explanatory. I will remove it.

neo/public/stylesheets/main.css
HotLemonJuice: It's a way to specify the behavior when the mouse hovers over the DOM. In this case, my purpose was to avoid the count div from blocking the mouse dragging the map. But I just found a bug, I will fix it.

HotLemonJuice: more explanation on the pointer-events: none: In addition to indicating that the element is not the target of mouse events, the value none instructs the mouse event to go "through" the element and target whatever is "underneath" that element instead.

JavierJia: good to know! thanks!

neo/public/stylesheets/main.css
```diff
@@ -131,22 +131,35 @@ search-bar {
...
   color: #000000;
-  margin-bottom: 30px;
+  margin-bottom: 40px;
   font-weight:500;
   float: left;
+  margin-top: 20px;
+  padding-right:20px;
+}
+
+#count-div > p{
```
JavierJia: the `> p` means select all `p` under `count-div`?

HotLemonJuice: yes, I wanted remove all of their margin-bottom to make two lines of text closer.

JavierJia:
from Chen's comment: align two count numbers and we can merge

HotLemonJuice:
## This is how it will look like
screen shot 2017-03-03 at 19 49 13
screen shot 2017-03-03 at 19 49 23

waans11:
Why don't we move "of" to the top right since "tweets" already occupies the space like the following?
       100,000 of
247,000,000 tweets
Personally, I think the first number is a little bit large.

chenlica:
My only preference is to align the two numbers on the right.  It's OK if we use the same size for both.

HotLemonJuice:
## New screenshots
screen shot 2017-03-04 at 12 58 30
screen shot 2017-03-04 at 12 54 34
screen shot 2017-03-04 at 12 59 15
screen shot 2017-03-04 at 13 00 23

HotLemonJuice:
Actually by following Taewoo's suggestions, I was able to use less HTML and CSS code.

JavierJia:
Looks good! 👍

HotLemonJuice:
New Screenshots:
screen shot 2017-03-04 at 13 22 50

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -34,7 +34,7 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
     countDiv.id = "count-div";
     countDiv.title = "Display the count information of Tweets";
     countDiv.innerHTML = [
-      "<p id='count'>{{ currentTweetCount | number:0 }}<span id='count-text'>&nbsp;&nbsp;of&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span></p>",
+      "<p id='count'>{{ currentTweetCount | number:0 }}<span id='count-text'>&nbsp;&nbsp;of&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span></p>",
```
JavierJia: I just noticed you are using space, right?
then what if the current count is 1 and the total count is 1,000,000,000 ?

JavierJia: I have a feeling that this align should be implemented by css code. e.g. put a table 
----------------
|    0    | of        |
| 1000 | tweets|
---------------
 and adjust the cell alignment?

HotLemonJuice: It will look like this. I used these spaces to fill the gap. E.g. "of      " -> "tweets"
screen shot 2017-03-04 at 13 26 43

JavierJia: interesting! then why `1` and `1,000,000,000` get aligned? curiously to know 😄

neo/public/stylesheets/main.css
```diff
@@ -128,11 +123,37 @@ search-bar {
...
+  margin-top: 20px;
+  margin-bottom: 40px;
+  padding-left: 10px;
+  float: left;
+}
+
+#count{
+  font-size: 32px;
+  margin-bottom: 0;
+  text-align: right;
```
JavierJia: aha, I found it. My bad! Good to go 😄

chenlica:
I agree with @JavierJia that it's better to use a mini table to do the alignment properly.

neo/public/stylesheets/main.css
JavierJia: @chenlica I found @HotLemonJuice specify the alignment in this CSS code without using a mini table. I think it's good enough. What's your opinion on this?

chenlica:
@JavierJia As long as the alignment is correct, we don't have to use a mini table.

JavierJia:
@HotLemonJuice You can go ahead and merge 👍

chenlica:
@HotLemonJuice and @JavierJia: When we load the page initially, it shows "0 of ...", which looks weird.  See the picture below.

Can we get rid of "0 of" when we first load the page, then show the real number of tweets in the results after a user submits a query?


screen shot 2017-03-06 at 9 29 53 am

JavierJia:
@HotLemonJuice wanna have to try?

HotLemonJuice:
OK, I will give it a try

Issue #254, opened 2017-02-23
Adjust the timeline position on the TwitterMap


JavierJia:
The timeline component is blocked if the resolution is lower. 
We should adjust the timeline width to a relative number to show it on a smaller resolution screen.

A quick way to debug the behavior is to use the Chrome dev tool to [adjust the webpage size](https://developers.google.com/web/tools/chrome-devtools/device-mode/emulate-mobile-viewports).

HotLemonJuice:
okay, I'm on it.

PR #259, opened 2017-02-24
[issue #254] [WIP] Adjust Timeline Width

HotLemonJuice:
related issue #254 
Purpose:
- Adjust the timeline width to make it fit in small screens.

Implementation Details:
- In the timeline/controller.js, set the timeline's width equal to the window size before drawing it.

# Screen Recordings: 
If I make it responsive to resizing event: redraw the timeseries everytime the window is resized, the web page will become very slow as shown in the recordings.
resize1 mov

resize12 mov


neo/public/javascripts/timeseries/controllers.js
```diff
@@ -39,6 +40,16 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
         }
       }
     );
+
+    console.log("initial width: " + $scope.responsiveWidth);
```
JavierJia: we should remove console logs

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -39,6 +40,16 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
...
       }
     );
+
+    console.log("initial width: " + $scope.responsiveWidth);
+    angular.element($window).bind('resize', function() {
+      $scope.responsiveWidth = $window.innerWidth;
+      // manuall $digest required as resize event
+      // is outside of angular
+      $scope.$digest();
+      console.log("width changed: " + $scope.responsiveWidth);
```
JavierJia: we should remove console logs

JavierJia:
@HotLemonJuice The previous comments has been unaddressed for 5 days.... Could you make a quick fix?

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: got it

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: Hi Jianfeng, this part of code was intended for making the timeseries responsive to resizing event. I think making the width responsive to the initial window size is not good enough. Let me work on making it responsive to resizing event.

HotLemonJuice:
- Current implementations will allow the timeline's width to be the width of the window when the page is loaded. So if the user resizes the window and find the width wrong, he or she can just refresh the browser. This implementation has no impact on the performance. 

- Another implementation is to redraw the timeline whenever the window is resized, but it will make the web page very laggy. There maybe some way of hacking the problem to resolve the performance issue. But I wonder if this implementation is really necessary. Any comments or suggestions?

neo/public/javascripts/timeseries/controllers.js
```diff
@@ -7,6 +7,7 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
     $scope.dc = $window.dc;
     $scope.crossfilter = $window.crossfilter;
     $scope.empty = [];
+    $scope.responsiveWidth = $(window).width();
```
JavierJia: questions, what is this `responsiveWidth`?

JavierJia:
I would pick the first one 😄

neo/public/javascripts/timeseries/controllers.js
HotLemonJuice: This name was intended for the responsive design: meaning the width can adjust itself according to the resizing event. I will change it now as we don't need that much responsiveness.

HotLemonJuice:
@JavierJia So is this PR good to merge?

chenlica:
I also prefer the first one.  If @JavierJia approves this PR, please do the merge.

JavierJia:
I will give you a plus 👍 
but the code is out-of-date, you should update the branch first.

PR #261, opened 2017-02-27
[Issue#235] Support the fast and estimate result in BaseDataAgent

JavierJia:
This PR add an interface to answer an estimable query using stats information.
Major changes:

Cloudberry

- Add an `isEstimable` field to the `Query` object
- Split the unified the `DataSetAgent` to `MetaDataAgent`, `ViewDataAgent`, and `BaseDataAgent` to clearly separate different kinds of workloads
- The `BaseDataAgent` collects the stats information periodically which will be used to estimate the answer to an `estimable` query. Current `stats` implementation is to only collect the total count.

`Neo` app 

- `NeoActor` ask for count query to `Berry` via HttpRequest every second.

Though it looks big, test cases contribute most of the codes.

Issue #235


neo/app/actor/NeoActor.scala
```diff
@@ -2,74 +2,95 @@ package actor
...
+          handleSamplingResponse(berryRequest)
+          handleSlicingResponse(berryRequest)
+        }.recoverTotal {
+          e =>
+            out ! JsError.toJson(e)
+        }
       }
+    case r: TotalCountRequest =>
+      handleCountResponse(r.berryJson)
+    case x => log.error("unknown:" + x)
```
haochen07: I am not sure if `log.error` will work. `Logger.logger.error` should do.

neo/app/actor/NeoActor.scala
```diff
@@ -2,74 +2,95 @@ package actor
...
         }
         res.body.via(Framing.delimiter(ByteString("\n"),
-                     maximumFrameLength = config.MaxFrameLengthForNeoWS,
-                     allowTruncation = true))
-                .runWith(sink)
-                .onFailure { case e => Logger.logger.error("NeoActor websocket receiving ... " + e.getMessage) }
+          maximumFrameLength = config.MaxFrameLengthForNeoWS,
+          allowTruncation = true))
+          .runWith(sink)
+          .onFailure { case e => Logger.error("NeoActor websocket receiving ... " + e.getMessage) }
```
haochen07: same as above

neo/app/actor/NeoActor.scala
```diff
@@ -2,74 +2,95 @@ package actor
...
       } else {
-        Logger.logger.error("Bad Gate Way. Connection code: " + res.headers.status)
+        Logger.error("Bad Gate Way. Connection code: " + res.headers.status)
```
haochen07: same as above

neo/app/actor/NeoActor.scala
JavierJia: Changed to `Logger.error`. 
I found previously we are using `play.api.Logger` which won't work. We actually should use `play.Logger` which showed the expected log. Don't know why they have two identical things.

neo/app/actor/NeoActor.scala
JavierJia: we are using `play.Logger` now 😄

neo/app/actor/NeoActor.scala
JavierJia: ditto

neo/app/actor/NeoActor.scala
haochen07: Cool. That is the thing we've been looking for for quite a while.

JavierJia:
@haochen07 usually we would say `Looks good` or `Looks bad` to merge 😁 . And if you approval the merge you can reply `:+1:`  👍  which is a clear sign of `go ahead and merge`

haochen07:
Looks good! 👍

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BaseDataAgent.scala
```diff
@@ -0,0 +1,116 @@
...
+import akka.pattern.pipe
+import edu.uci.ics.cloudberry.zion.common.Config
+import edu.uci.ics.cloudberry.zion.model.datastore.{IDataConn, IQLGenerator}
+import edu.uci.ics.cloudberry.zion.model.schema._
+import org.joda.time.{DateTime, Duration}
+import play.api.libs.json._
+
+import scala.concurrent.ExecutionContext
+
+class BaseDataAgent(override val dbName: String,
```
luochen01: The name of "BaseDataAgent" is a little bit confusing with "AbstractDataSetAgent", since base and abstract are sometimes both used for "parent/super" class in OOP. I would suggest using a different name for "BaseDataAgent", such as "DataAgent"?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -33,7 +33,7 @@ class DataStoreManager(metaDataset: String,
   val managerParser = queryGenFactory()
   implicit val askTimeOut: Timeout = Timeout(config.DataManagerAppendViewTimeOut)
 
-  val metaActor = childMaker(context, "meta", Seq(DataSetInfo.MetaSchema, queryGenFactory(), conn, ec))
+  val metaActor: ActorRef = childMaker(AgentType.Meta, context, "meta", DataSetInfo.MetaDataDBName, DataSetInfo.MetaSchema, queryGenFactory(), conn, config)
 
   override def preStart(): Unit = {
     metaActor ? Query(metaDataset, select = Some(SelectStatement(Seq.empty, 100000000, 0, Seq.empty))) map {
```
luochen01: Why there is a magic number here? Would it better be moved to a centralized place for ease of configuration?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BaseDataAgent.scala
JavierJia: Good suggestion. Renamed to `OriginalDataAgent`.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
JavierJia: hmm, good point. this number is rather a `max` number than a `magic` number. it intends to load all. but since our grammar forced to have the limit, so we put a big enough number there. 

Changed to `Integer.Max` to be less magic.

JavierJia:
One small suggestion for reviewers, I think there are two buttons (a)`send a single review` and (b) `start a new review` (not the exact words). I would suggest use (b) to send the review comments in a batch way to reduce the email notification overhead. 😃

luochen01:
Looks good! 👍

PR #264, opened 2017-03-01
Close the websocket when Neo get killed

JavierJia:
Since the web socket will keep alive for quite a long time, the berry actor will keep working on the queries even the client is closed. This PR fix this issue.

### DIFF
--- a/neo/app/controllers/Application.scala
+++ b/neo/app/controllers/Application.scala
@@ -7,20 +7,20 @@ import actor.NeoActor
 import akka.actor.{Actor, ActorLogging, ActorRef, ActorRefFactory, ActorSystem, DeadLetter, OneForOneStrategy, PoisonPill, Props, Status, SupervisorStrategy, Terminated}
 import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
 import akka.stream.{Materializer, OverflowStrategy}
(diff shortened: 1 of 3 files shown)


neo/app/controllers/Application.scala
```diff
@@ -41,7 +41,7 @@ class Application @Inject()(val wsClient: WSClient,
 
   val manager = system.actorOf(DataStoreManager.props(Migration_20160814.berryMeta, asterixConn, AQLGenerator, config))
 
-  Logger.logger.info("I'm initializing")
+  Logger.info("I'm initializing")
```
haochen07: I also tried this on my local version, `Logger.info` will not print the information on console. `Logger.error` works, `Logger.logger.info` works. Maybe something related to log configuration?

JavierJia: That would be weird. I indeed saw them in the `logs/application.log`. I think it should be something wrong with our logback.xml settings. You can have a try to see if what is a good setting 🤔

JavierJia:
I forget to try multiple uses, it actually closes **ALL** clients. So it's not the correct way to disconnect the ws.

neo/app/controllers/Application.scala
haochen07: OK I'll check that

PR #266, opened 2017-03-02
[Issue#256] Implement SQL++ Query Generator

luochen01:
This PR adds the implementation of SQL++ query generator, which generates SQL++ queries for Cloudberry queries. It also supports lookup multiple datasets/multiple select values mentioned in Issue#194. It also fixes a bug in Issue#255, where the AQLGenerator ignores the unnested variable when it's not explicitly selected.

Major changes:

- Added `QueryResolver`, which resolves referenced fields in `Query`, which can be used during the query generation phase
- Added `QueryValidator`, which validates the correctness of `Query`
- Added some fields in `Query`, which stores the referenced fields after `QueryResolver`
- Changed some variable names of `Query` to make them more meaningful
- Added `AbstractQueryGenerator`, which provides basic dispatch functions and calls `QueryResolver` and `QueryValidator` for query analysis
- Added `AsterixQueryGenerator`, which provides some common basic functions for AQL/SQL++ generators
- Added `SQLPPGenerator`, which generates SQL++ queries
- Added test cases for `SQLPPGenerator`, and modified test cases for `AQLPPGenerator`
- Modified `AQLGenerator` to be consistent with the current class hierarchy. Also fixed the bug #266 and provides support lookup multiple datasets/select values #194 
- Removed `AsterixFuncVisitor`, which is replaced by `AsterixQueryGenerator`


zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Query.scala
```diff
@@ -27,29 +28,41 @@ object QueryExeOption {
...
-                 groups: Option[GroupStatement] = None,
+                 lookups: Seq[LookupStatement] = Seq.empty,
+                 filters: Seq[FilterStatement] = Seq.empty,
+                 unnests: Seq[UnnestStatement] = Seq.empty,
+                 group: Option[GroupStatement] = None,
                  select: Option[SelectStatement] = None,
                  globalAggr: Option[GlobalAggregateStatement] = None,
                  isEstimable: Boolean = false
                 ) extends IReadQuery {
+  var fieldMap: Map[String, Field] = null
```
JavierJia: we should rarely use `var`, especially for the query object.

I assume all these fields are used for `QLGenerator`. If that is the case, it's better to make a separate object, e.g., `QueryFields`, along with the immutable `Query` object. The rest of the codebase won't care about these `QueryFields`. 

Even in that case, since they are all `Map`, it's will be better to use `val fieldX: mutable.Map`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/QueryResolver.scala
```diff
@@ -0,0 +1,293 @@
...
+
+import edu.uci.ics.cloudberry.zion.model.datastore.QueryValidator._
+import edu.uci.ics.cloudberry.zion.model.schema.Relation._
+import edu.uci.ics.cloudberry.zion.model.schema._
+
+import scala.collection.mutable
+import scala.reflect.runtime.universe.{TypeTag, typeOf}
+
+
+object QueryValidator {
```
JavierJia: conventionally, we put the class object at first, the companion object at last.

chenlica:
Great progress!  Is it possible to split it into smaller PRs?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Query.scala
luochen01: Got it. I'll extract those fieldMaps into a separate class.

luochen01:
Not sure about that. All changes are somewhat related, and not sure whether the project can compile properly if the commits are submitted alone. But I'll make small PRs next time step by step...

luochen01:
I submitted a few commits to
1. reverse some variable name changes
2. reverse code changes due to formatting issue (I thought the default code format of IntelliJ will have the exact same configuration, but it turns out it's not...sorry for that)

Now the changes are mainly limited to `zion.model.impl` package, which should be expected.

Issue #265, opened 2017-03-01
Neo should cancel the previous query if new query comes


JavierJia:
Previously, we are using one BerryClient to answer one specific NeoActor. In that case, whenever there is a new request, the berryClient will return to the initial status. E.g., a user may ask "zika", before the entire timeline finishes, he may get bored and ask a new one, say "flood". The "zika" related result won't be returned.

However, since we are using the web socket connection with BerryClient, each query will be connected with different BerryClient!! So the previous working BerryClient won't notice that the request has changed. It will keep sending the old result. As the result, for the aforementioned scenario, the current online demo will return the mixed results for both queries❗️ 

A simple fix might be canceling the previous `Batch` request. Some examples from StackOverflow: http://stackoverflow.com/questions/16009837/how-to-cancel-future-in-scala

A more sophisticated solution is that we should design a `cancel` protocol between clients and berryClient.

chenlica:
Who should be the right person to do the fix?  @haochen07 ?

JavierJia:
Yes, I would encourage Hao to have a try 😄

haochen07:
Ok I'll try to fix that

JavierJia:
As we discussed today, we will directly remove the `NeoActor` and only use the `ws` API to communicate with front-end.

haochen07:
OK

PR #269, opened 2017-03-05
[Issue #265] Remove NeoActor and introduce RequestRouter to differentiate streaming and non-streaming requests

haochen07:
**[Not totally finished]**
This PR is to fix problems in #265 and involves **architecture changes**.
(_The major part and logics are finished in this PR. Details need to be discussed and polished._)

**Problem**
The cause of the problem is that we are using web socket connection with `BerryClient`, each query will be dealt with different client. The working client will not notice that the query has changed and continue sending results.

**Solution**
- The new solution changes the logic of interactions between front-end and `BerryClient`. 
- Generally, `NeoActor` is responsible for (1) generating berry requests and (2) wrapping berry response. I removed `NeoActor` and let front-end deal with (1) and middleware deal with (2). A new actor `RequestRouter` is introduced in the entrance of middleware, responsible for differentiating non-streaming and streaming requests and assign proper `BerryClient` to deal the task. The front-end is able to communicate directly to middleware without a third agent.
- This is for now the ultimate solution because (1) solving the mentioned problem by cause, (2) streaming and non-streaming requests are treated separately, good for further functionalities, (3) make front-end and middleware independent logically.

**Implementation**
- Generated `berryRequest` instead of `UserRequest` in front-end.
- Designed the request and response format into a more general version. 
- Built `RequestRouter` to distribute streaming request to one `berryClient` and non-streaming request to another, and transform berry response into general format and send back to front-end.
- The front-end is now responsible for sending total count requests.
- Error messages are logged out in server side and also returned to frone-end's `errorMessage`.
- Removed `NeoActor` and `NeoActorTest`.

**To be done**
- to catch streaming response and return to front-end.
- test case of `RequestRouter`

**Comments**
1. The `UserRequest` model is now of no use since we totally change the format. Shall we remove it?
2. Do we still need the API for `Berry`, now `RequestRouter` is the only entrance for them.
3. Need comments on the mechanism of error treatment.


JavierJia:
once we finished on task, usually we will remove the feature branch, sync with the upstream master, and then checkout a new branch from the latest master. In this way, your commit history will be much clean.

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,107 @@
...
+import edu.uci.ics.cloudberry.zion.common.Config
+import play.api.libs.json._
+import play.api.libs.ws.WSClient
+import play.api.mvc.RequestHeader
+import play.Logger
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext
+
+class RequestRouter (out: ActorRef, ws: WSClient, requestHeader: RequestHeader, berryClientProp: Props, config: Config)
```
JavierJia: `ws` is not needed, right?

JavierJia: neither `requestHeader` ?

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,107 @@
...
+                    (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
+
+  import RequestRouter._
+  import KeyType._
+
+  implicit val timeout: Timeout = Timeout(20.minutes)
+  val StreamingBerryClient = context.actorOf(berryClientProp)
```
JavierJia: if it's not global value, the first letter should be lowercase.

haochen07:
Got it.

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,107 @@
...
+  val nonStreamingBerryClient = context.actorOf(berryClientProp)
+
+  // TODO UserRequest in package model is of no use
+  override def receive: Receive = {
+    case requestBody: JsValue =>
+      (requestBody \ "transform" \\ "key")(0).asOpt[KeyType.Value] match {
+        case Some(TotalCount) =>
```
JavierJia: I think `TotalCount` and the `Sample` is the value of the `key` field. As a general purpose request handler, it shouldn't handle the user's logic.

I think my previous idea of simplification in Slack is not precise. My bad 😭 

The correct way should be something like following
```
 case body : JsValue =>
    val transform = parseTransform(body)
    (body \\ "slicing") match {
          case Some(x) => handleStreaming(x, transform)
          case None => handleNonStreaming(body, transform)
     }
...

// here is the simplification I meant yesterday.
def parseTransform(body: JsValue) = {
   (body \\ "wrapper").asOpt[JsValue] match {
        case None => Transform.NonTransform
        case Some(x) => Transform.Wrapper( (x \ "key").as[String], (x \ "value").as[String]) ... 
   } 
}

trait Transform = {
   def tranform(jsvalue: JsValue): JsValue
}
object Transfrom {
  val NoTransform = new Transform ...
  val WraperTransform(key: String, value: String) = new Transform ....
}
```

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+}
+
+object RequestRouter {
+
+  def props(out: ActorRef, ws: WSClient, requestHeader: RequestHeader, berryClientProp: Props, config: Config)
+           (implicit ec: ExecutionContext, materializer: Materializer) = Props(new RequestRouter(out, ws, requestHeader, berryClientProp, config))
+
+  object KeyType extends Enumeration {
```
JavierJia: As a general purpose middleware, it shouldn't have these enum that only make sense for TwitterMap

neo/app/controllers/Application.scala
```diff
@@ -59,7 +60,8 @@ class Application @Inject()(val wsClient: WSClient,
   }
 
   def ws = WebSocket.accept[JsValue, JsValue] { request =>
-    ActorFlow.actorRef(out => NeoActor.props(out, wsClient, request.host, config))
+    val prop = BerryClient.props(new JSONParser(), manager, new QueryPlanner(), config)
```
JavierJia: 👍

neo/app/controllers/Application.scala
```diff
@@ -69,6 +71,7 @@ class Application @Inject()(val wsClient: WSClient,
     }
   }
 
+  // TODO delete this??
```
JavierJia: NONONONO.

We should keep this one. It will be useful for people that are using HTTP. We need to provide more logical protocol for them. (In the future).

neo/public/javascripts/common/services.js
```diff
@@ -1,12 +1,37 @@
...
+    var defaultNonSamplingDayRange = 1500;
+    var defaultSamplingDayRange = 1;
+    var defaultSamplingSize = 10;
     var ws = new WebSocket("ws://" + $location.host() + ":" + $location.port() + "/ws");
 
-    ws.onopen = function() {
-      var json = JSON.stringify({ cmd : "totalCount"});
-      ws.send(json);
-    };
+    var countRequest = JSON.stringify({
```
JavierJia: I see, here is the place that causes the confusion.
The request should be like this
```
    transform: {
         wrap: {
                  key : "totalCount",
                  value : "value"  // means the returned result is in the `value` field
         }
     }
     // all the original fields
 
     dataset: "twitter.ds_tweet",
      global: {
            globalAggregate: {
               field: "*",
               apply: {
                 name: "count"
                },
                as: "count"
           }},
           estimable : true
          }
        }

```

neo/public/javascripts/common/services.js
```diff
@@ -32,50 +57,183 @@ angular.module('cloudberry.common', [])
...
-          case "byTime":
-            asterixService.timeResult = result.value;
-            break;
-          case "byHashTag":
-            asterixService.hashTagResult = result.value;
-            break;
+        switch (result.transform.wrap.key) {
           case "sample":
-            asterixService.tweetResult = result.value[0];
+            asterixService.tweetResult = result.transform.wrap.value[0];
```
JavierJia: the result should not inside the `transform`, the result should be just one record
```
   {
       key : `sampel`, // or else
       value : [ ]    // a list of result
   }
```

JavierJia:
wait, you don't need to close it ...

JavierJia:
It just a minor suggestion. I don't want to waste your time on the cleaning 😄  can you reopen it?

haochen07:
Yes... but I have already deleted the previous branch to clean the history... I'll look at all the comments and fix them in a new PR.

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,107 @@
...
+          }.recover{
+            case e =>
+              val errorMsg = "Failed to receive response from non-streaming berryClient in Sample query: " + e.getMessage
+              Logger.error(errorMsg)
+              respondWrapper(JsString(errorMsg), Error)
+          }.pipeTo(out)
+
+        case Some(Batch) =>
+
+          // TODO need to implement
```
JavierJia: This just remind me why I designed the `transformation` function in Berry!!!
In order to transfer the streaming result, we cannot use the `ask` model, since it will just receive one result and close the channel. So, instead of handling the `transformation` logic in the `router`, we have to let `Berry` transform it and directly send to the client by using the `tell` (`!`) model.

JavierJia:
OK, no problems. I think the JS part is almost there, which is the most important one.
The actor part need to pick up the original `IPostTransform` function and let berry to handle the transform work.

neo/app/actor/RequestRouter.scala
haochen07: then where is the `slicing` field? 

```
transform:{
  wrap:{
    key: 'key',
    value: {
      
    }
    slicing: true or false
  }
}
```
like this?

chenlica:
An old question: can we split it into smaller PRs? :-)

JavierJia:
@chenlica the large chunks of commits dues to @haochen07 didn't checkout from the most recent master. It contains a lot of existing commits. Since this is an API-like change, the change set makes sense to me 😄 

@haochen07, I write a small instruction about to checking out a clean branch [here](https://github.com/ISG-ICS/cloudberry/wiki/Setting-up-the-development-environment#start-a-new-feature-branch). You can take a look for the future PR.

Issue #260, opened 2017-02-26
Separate the TwitterMap with Cloudberry Web Application


JavierJia:
Currently, they are mixed in the same object `Application.scala`. 
We can separate them by creating two `Applicatons`, e.g., `CloudberryApp` and `TwitterMapApp`. Each app has its own route entry to clearly separate the front-end apps with Cloudberry.

PR #270, opened 2017-03-06
[Issue #260 #265] Remove NeoActor and introduce RequestRouter to differentiate streaming and non-streaming requests

haochen07:
This PR is to fix problems in #265 #260 and involves **architecture changes**.

**Problem**
The cause of the problem is that we are using web socket connection with `BerryClient`, each query will be dealt with different client. The working client will not notice that the query has changed and continue sending results.

**Solution**
- The new solution changes the logic of interactions between front-end and `BerryClient`. 
- Generally, `NeoActor` is responsible for (1) generating berry requests and (2) wrapping berry response. I removed `NeoActor` and let front-end deal with (1) and middleware deal with (2). A new actor `RequestRouter` is introduced in the entrance of middleware, responsible for differentiating non-streaming and streaming requests and assign proper `BerryClient` to deal the task. The front-end is able to communicate directly to middleware without a third agent.
- This is for now the ultimate solution because (1) solving the mentioned problem by cause, (2) streaming and non-streaming requests are treated separately, good for further functionalities, (3) make front-end and middleware independent logically.

**Implementation**
- Generated `berryRequest` instead of `UserRequest` in front-end.
- Designed the request and response format into a more general version. 
- Built `RequestRouter` to distribute streaming request to one `berryClient` and non-streaming request to another, and transform berry response into general format and send back to front-end.
- The front-end is now responsible for sending total count requests once per second.
- Error messages are logged out in server side.
- Fixed the `logger` configuration problem.
- Removed `NeoActor`, `NeoActorTest`, `UserRequest` and `UserRequestTest`, as well as all configurations related to `NeoActor`.


haochen07:
Sorry I tried several times but I failed to remove all commit history...
Need suggestions on:
1. The data model `UserRequest` is of no use now, since front-end send berry request directly. Shall we remove it?
2. Error messages are logged out in server side. Do we need to forward error messages to front-end?

JavierJia:
> The data model UserRequest is of no use now, since front-end send berry request directly. Shall we remove it?

Yes.

> Error messages are logged out in server side. Do we need to forward error messages to front-end?

Yes, but in a separate PR and lower priority.

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,72 @@
...
+  override def receive: Receive = {
+    case requestBody: JsValue =>
+      val transformer = parseTransform(requestBody)
+      val originRequestBody = getOriginRequestBody(requestBody)
+      (originRequestBody \\ "sliceMillis").isEmpty match {
+        case true => handleNonStreamingBody(originRequestBody, transformer)
+        case false => handleStreamingBody(originRequestBody, transformer)
+      }
+    case e =>
+      // TODO: Do we need to return error to front-End ??
```
JavierJia: we should but in a separate PR.

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,72 @@
...
+      case Some(t) =>
+        (t \ "wrap").asOpt[JsValue] match {
+          case Some(w) => WrapTransform((w \"key").as[String])
+          case None => NoTransform
+        }
+      case None => NoTransform
+    }
+  }
+
+  private def getOriginRequestBody(requestBody: JsValue): JsValue = {
```
JavierJia: `getOriginRequestBody` --> `getBerryRequest` ?

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import scala.concurrent.ExecutionContext
+
+class RequestRouter (out: ActorRef, berryClientProp: Props, config: Config)
+                    (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
+
+  import RequestRouter._
+
+  val streamingBerryClient = context.actorOf(berryClientProp)
+  val nonStreamingBerryClient = context.actorOf(berryClientProp)
```
JavierJia: berryProp should use `out`, right? otherwise, how could berry send the msg to?

neo/app/actor/RequestRouter.scala
```diff
@@ -0,0 +1,72 @@
...
+
+}
+
+object RequestRouter {
+  def props(out: ActorRef, berryClientProp: Props, config: Config)
+           (implicit ec: ExecutionContext, materializer: Materializer) = Props(new RequestRouter(out, berryClientProp, config))
+
+  case class WrapTransform(key: String) extends IPostTransform {
+    override def transform(jsonBody: JsValue): JsValue = {
+      Json.obj("key" -> key, "value" -> jsonBody)
```
JavierJia: 👍

neo/app/controllers/Application.scala
```diff
@@ -59,7 +60,8 @@ class Application @Inject()(val wsClient: WSClient,
   }
 
   def ws = WebSocket.accept[JsValue, JsValue] { request =>
-    ActorFlow.actorRef(out => NeoActor.props(out, wsClient, request.host, config))
+    val prop = BerryClient.props(new JSONParser(), manager, new QueryPlanner(), config)
+    ActorFlow.actorRef(out => RequestRouter.props(out, prop, config))
```
JavierJia: I think berry prop should also be a function of `out`.

neo/app/actor/RequestRouter.scala
JavierJia: we should either pass a function or directly create the `berry` inside.

pass a function 
```
class RequestRouter (out: ActorRef, berryClientProp: ActorRef => Props, config: Config)
          (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
    val streamingBerryClient = context.actorOf(berryClientProp(out))
```

neo/public/javascripts/common/services.js
```diff
@@ -1,12 +1,36 @@
...
+          apply: {
+            name: "count"
+          },
+          as: "count"
+        }},
+      estimable : true,
+      transform: {
+        wrap: {
+          key: "totalCount",
+          value: {}
```
JavierJia: since the `value` is not used, we can just remove this field.

neo/public/javascripts/common/services.js
```diff
@@ -32,34 +56,167 @@ angular.module('cloudberry.common', [])
...
+            offset: 0,
+            field: ["create_at", "id", "user.id"]
           },
-          timeBin : parameters.timeBin,
-          geoLevel: parameters.geoLevel,
-          geoIds : parameters.geoIds
+          transform: {
+            wrap: {
+              key: "sample",
+              value: {}
```
JavierJia: ditto.

neo/public/javascripts/common/services.js
```diff
@@ -32,34 +56,167 @@ angular.module('cloudberry.common', [])
...
+
+        var batchJson = (JSON.stringify({
+          batch: [this.byTimeRequest(parameters), this.byGeoRequest(parameters), this.byHashTagRequest(parameters)],
+          option: {
+            sliceMillis: 2000
+          },
+          transform: {
+            wrap: {
+              key: "batch",
+              value: {}
```
JavierJia: ditto

neo/public/javascripts/common/services.js
```diff
@@ -32,34 +56,167 @@ angular.module('cloudberry.common', [])
...
+              apply: {
+                name: "count"
+              },
+              as: "count"
+            }]
+          }
+        };
+      },
+
+      byTimeRequest: function(parameters) {
```
JavierJia: I think all these js functions should be outside the `var asterixService `

neo/app/controllers/Application.scala
haochen07: If we give `BerryClient` an `out` in `prop`, it will send the response directly to `out`, not the `sender()`, namely RequestRouter.

neo/app/actor/RequestRouter.scala
haochen07: `Berry` will send to `out`, which is defined in `tell`.

neo/app/controllers/Application.scala
JavierJia: correct. since `berry` knows how to ship the result, it can directly send to out. 
otherwise, the `RequestRouter` has to receive the result and redirect it to `out` which is not reflect in the code. AND, it's not necessary.

neo/app/actor/RequestRouter.scala
JavierJia: I see. 
in that case, you can use `forward` function to be clear 😄 
But, as explained earlier, we'd better to clearly deliver the msg by directly give out to berry.

neo/app/controllers/Application.scala
haochen07: So we pass `out` into `Berry` and in `RequestRouter`, we only send the request to `berry` and let `berry` itself deal with sending back the response?

JavierJia: I see it now 😃

JavierJia: correct 👍

JavierJia: one correction, the flow is always `Router` -> `Berry` -> `Out`. 
What I mean is to make the code clearly deliver this message by build the Berry with this `out`.

haochen07: Ok

haochen07:
- pass `out` to `BerryClient.prop`
- Router is responsible for sending request and `berry` is responsible for sending back results.
- Refactor front-end function style.
- Remove `UserRequest` and its test.

haochen07:
In test case, I defined a client to return a result per 500 millis if a slicing request is received. I sent the client one slicing request and interrupted it by sending another slicing request. However, it seems that the client still returns the full results.
Is there something wrong with the test logic? or something wrong with router implementation?

neo/test/actor/RequestRouterTest.scala
```diff
@@ -1,30 +1,221 @@
...
+        "value" -> berryResult
+      ))
+
+      ok
+    }
+
+    "distribute non-slicing and slicing requests to different berry clients" in {
+
+      val nonSlicingBerryResult = Json.obj("result" -> "non slicing result")
+      val slicingBerryResultOne = Json.obj("result" -> "slicing result 1")
```
JavierJia: let's call it `streaming` to be clear?

neo/test/actor/RequestRouterTest.scala
```diff
@@ -1,30 +1,221 @@
...
+      frontEnd.expectMsg(slicingBerryResultOne)
+      frontEnd.expectMsg(slicingBerryResultTwo)
+      frontEnd.expectMsg(slicingBerryResultThree)
+
+      frontEnd.send(router, slicingRequestWithNoTransform)
+      frontEnd.expectMsg(100 millis, slicingBerryResultOne)
+      frontEnd.send(router, slicingRequestWithNoTransform)
+
+      /*
+      ideal test case result:
```
JavierJia: because the `clientProps` that you defined is not as the same as the `berryClient`. The `berryClient` itself will cancel the previous query.  It's been tested. So for here, we need to produce a real `berryClient`, and only mock the `ws` connection. However, it will be an integration test not a unit test. For this `router`, as long as the msg is delivered to berry it should be good.

neo/test/actor/RequestRouterTest.scala
```diff
@@ -1,30 +1,221 @@
...
+      val frontEnd = new TestProbe(system)
+      val clientProps = Props(new Actor {
+        override def receive: Receive = {
+          case berryRequest: Request =>
+            frontEnd.ref ! berryRequest.postTransform.transform(berryResult)
+          case _ =>
+            throw new IllegalArgumentException("Message type sent to berry is not correct.")
+        }
+      })
+      val router = system.actorOf(RequestRouter.props(frontEnd.ref, clientProps, Config.Default))
```
JavierJia: I think this test runs a little far. In the perspective of `router`, as long as it can route the result to different clients that will be enough. 

But I found there are some difficulties to simplify the test logic, which usually means the API we have designed should be improved. Let me think about it.

JavierJia: One solution to that to make a `childmaker` as we did in `DataStoreManager`. Instead of giving `router` a  specific `prop`, we directly give it a function that can create the `ActorRef`.
Then in the source code, we can implement the `childMaker` using `prop`, and in the test, we just passing already existed `ActorRef`s.

JavierJia: since the `clientProps` will create two different actors inside, then the `frontEnd` should receive two kinds of msgs from different actors, not the same one, then if all msgs are complete, and different msgs are coming from two different senders, it serves the purpose of testing the `router`.

haochen07:
I came up with a new idea that
in `application`
```
val streamingProp = berry.prop(..., name1)
val nonStreamingProp = berry.prop(..., name2)
RequestRouter.prop(..., streamingProp, nonStreamingProp)
```
and in `router`
```
val streamingClient = context.actorOf(streamingProp)
val nonStreamingClient = context.actorOf(nonStreamingProp)
```

neo/test/actor/RequestRouterTest.scala
haochen07: I came up with the idea that
in `application`
```
val streamingProp = berry.prop(..., name1)
val nonStreamingProp = berry.prop(..., name2)
RequestRouter.prop(..., streamingProp, nonStreamingProp)
```
and in `router`
```
val streamingClient = context.actorOf(streamingProp)
val nonStreamingClient = context.actorOf(nonStreamingProp)
```

haochen07: And in `test`, assign the actor name in the response. Is that work?

haochen07: Or we don't need to change the `router` implementation. Use `self.path.name` to return actors' names to recognize them.

JavierJia: yes, it will work!

JavierJia: clean and simple! 👍

JavierJia: make sure, the 
```
val streamingProp = berry.prop(..., name1)
val nonStreamingProp = berry.prop(..., name2)
RequestRouter.prop(..., streamingProp, nonStreamingProp)
```
 is inside the `ws` function not the global variables.

haochen07: Of course.

haochen07: Actually it is even simpler than that. The name is assigned in `context.actorOf`, which I gave each of them different names. See the new commit.

neo/app/controllers/Application.scala
```diff
@@ -59,7 +60,9 @@ class Application @Inject()(val wsClient: WSClient,
   }
 
   def ws = WebSocket.accept[JsValue, JsValue] { request =>
-    ActorFlow.actorRef(out => NeoActor.props(out, wsClient, request.host, config))
+    ActorFlow.actorRef{ out =>
+      RequestRouter.props(out, BerryClient.props(new JSONParser(), manager, new QueryPlanner(), config, out), config)
```
JavierJia: do we need `out` for `router`?

neo/public/javascripts/common/services.js
```diff
@@ -68,18 +222,18 @@ angular.module('cloudberry.common', [])
             asterixService.mapResult = result.value[1];
             asterixService.hashTagResult = result.value[2];
             break;
+          case "totalCount":
+            asterixService.totalCount = result.value[0][0].count;
+            break;
           case "error":
-            console.error(result);
+            console.error(result.toString());
```
JavierJia: in JS print the object is better. since you can drill down to each nested field like in the debug mode.

neo/public/javascripts/common/services.js
```diff
@@ -1,12 +1,157 @@
 angular.module('cloudberry.common', [])
   .service('Asterix', function($http, $timeout, $location) {
```
JavierJia: we've blaming for this `Asterix` name for a long time. could you fix it by this time by correctly name this service `Cloudberry`? thanks!

neo/public/javascripts/common/services.js
```diff
@@ -68,18 +222,18 @@ angular.module('cloudberry.common', [])
...
             break;
           case "done":
             break;
-          case "totalCount":
-            asterixService.totalCount = result.value[0][0].count;
-            break;
           default:
-            console.error("ws get unknown data:" );
-            console.error(result);
+            console.error("ws get unknown data: " + result.toString());
```
JavierJia: ditto

neo/public/javascripts/common/services.js
haochen07: which means just change the name from `Asterix` to `Cloudberry`?

JavierJia: yes, and may also need to clean the variable names?

haochen07: Changing the name will lead to a fail in front-end. There is no map appeared.
`$http` can be eliminated. But since I didn't find where the function is called, I am not sure if the removal will influence the performance. Based on my simple test on local machine, it is good.

JavierJia: Haha, that's fine. I'm asking too much. let's just keep this PR simple.

neo/app/controllers/Application.scala
```diff
@@ -41,7 +41,8 @@ class Application @Inject()(val wsClient: WSClient,
 
   val manager = system.actorOf(DataStoreManager.props(Migration_20160814.berryMeta, asterixConn, AQLGenerator, config))
 
-  Logger.logger.info("I'm initializing")
+  // TODO Logger.info cannot logout to console
+  Logger.info("I'm initializing")
```
haochen07: Fixed. Due to configuration at `logback.xml`. [Here](https://logback.qos.ch/manual/filters.html) is the reference. 
Now we can `log.info`, `log.warn`, `log.error` inside an actor that `with ActorLogging`, and `Logger.info`, `Logger.warn`, `Logger.error` in other classes. See the new commit.

haochen07:
I also removed `NeoActor` configurations.

neo/app/actor/RequestRouter.scala
```diff
@@ -1,21 +1,22 @@
...
 
-class RequestRouter (out: ActorRef, berryClientProp: Props, config: Config)
+class RequestRouter (berryClientProp: Props, config: Config)
                     (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
 
   import RequestRouter._
 
-  val streamingBerryClient = context.actorOf(berryClientProp, "streamingClient")
-  val nonStreamingBerryClient = context.actorOf(berryClientProp, "nonStreamingClient")
+  val streamingClientName = "streamingClient"
```
JavierJia: const values should go to `object RequestRouter`

neo/conf/logback.xml
```diff
@@ -10,12 +10,8 @@
   </appender>
 
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-    <filter class="ch.qos.logback.classic.filter.LevelFilter">
+    <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
```
JavierJia: cool!

neo/public/javascripts/common/services.js
```diff
@@ -226,13 +226,14 @@ angular.module('cloudberry.common', [])
...
-            console.error(result.toString());
+            console.error(result);
             asterixService.errorMessage = result.value;
             break;
           case "done":
             break;
           default:
-            console.error("ws get unknown data: " + result.toString());
+            console.error("ws get unknown data: ");
+            console.error(result);
```
JavierJia: JS is even simpler than scala 
`console.error("ws get unknown data: ", result);`

chenlica:
💯

Issue #157, opened 2016-08-08
Provide a point-view for the geo-location tagged tweets


JavierJia:
plot the tweets by it's geolocation as a point. 

It could be implemented by using the `geocell` function of the new data model.

JavierJia:
The aggregation result is easy to show. The harder problem is to show the information on data points.

JavierJia:
The "topK" may not serve the purpose directly. What we need is a "top10%" instead of "top100" result. Does *db support this query?

JavierJia:
@simonmssu

JavierJia:
Get an idea from a discussion in SIGSPATIAL: issue a aggregation result for each very small cell.

JavierJia:
@zonghengma we can do it step by step. 
The first step just show all points within a small region. e.g., when we zoom to a layer finner then city, we show all points in that region.

JavierJia:
@HotLemonJuice , could you *move* the issue content you just created to here? thanks!

HotLemonJuice:
So far we are using a heat map to show the Tweets on the map. It is a high-level description of the data.

Now if we can show the exact location of each Tweet, we can make the connection between the data and the geolocation closer.

A good example by MapD:
https://www.mapd.com/demos/tweetmap/
![screen shot 2017-03-06 at 16 41 41](https://cloud.githubusercontent.com/assets/12385178/23636764/389961c8-028c-11e7-92f1-d80b57434ac0.jpg)

However, since there are too much data to load in the browser, we need use some tricks to avoid crashing the browser. We can start implementing this feature in small areas of the map, e.g. city level.

HotLemonJuice:
There aren't much point data in the sample data. How do I add fake data into the DB?

JavierJia:
we can generate a sample data. for now, you may use the field of `"bounding_box": rectangle("-74.146932,40.643773 -74.0658,40.697794") }`?  you can just pick the first value, and add some random value.

JavierJia:
you can take a look at the `sample.data` under `script` folder. Here is an example of the `place` field
```
"place": { "country": "United States", "country_code": "United States", "full_name": "Houston, TX", "id": "1c69a67ad480e1b1", "name": "Houston", "place_type": "city", "bounding_box": rectangle("-95.823268,29.522325 -95.069705,30.154665") }
```

HotLemonJuice:
Notes for future dev:
[...]

PR #276, opened 2017-03-07
[issue #157] [WIP] Provide a point-view for the geo-location tagged Tweets

HotLemonJuice:
Issue #157 

Purpose:
- Plot the tweets by it's geolocation as a point.
screen shot 2017-03-06 at 16 41 41

Possible Implementation:
- the geocell function of the new data model.

### DIFF
--- a/script/generateSampleData.py
+++ b/script/generateSampleData.py
@@ -1,0 +1,22 @@
+import random
+
+text1 = '{ "create_at": datetime("2015-11-23T16:14:03.000Z"), "id": 668945640186101761, "text": "Just posted a photo @ Campus Martius Park https://t.co/5Ax4E2CdWZ", "in_reply_to_status": -1, "in_reply_to_user": -1, "favorite_count": 0, "coordinate": point("-83.04647491,42.33170228"), "retweet_count": 0, "lang": "en", "is_retweet": false, "user": { "id": 48121888, "name": "Kevin McKague", "screen_name": "KevinOfMI", "lang": "en", "location": "Davison, Michigan", "create_at": date("2009-06-17"), "description": "I need to ride my bike until my brain shuts up and my muscles are screaming. \\nRight after these donuts. Dad of 3.\\n Visit my blog 18 Wheels and a 12-Speed Bike.", "followers_count": 1178, "friends_count": 1780, "statues_count": 22263 }, "place": { "country": "United States", "country_code": "United States", "full_name": "Detroit, MI", "id": "b463d3bd6064861b", "name": "Detroit", "place_type": "city", "bounding_box": rectangle("'
+middleText = '-83.288056,42.255085 -82.91052,42.450488'
+text2 = '") }, "geo_tag": { "stateID": 26, "stateName": "Michigan", "countyID": 26163, "countyName": "Wayne", "cityID": 2622000, "cityName": "Detroit" } }'
+
+count = 200
+
+for i in range(count):
+    sign = random.randrange(0, 3, step=2) - 1
+    n = random.randrange(0,1000000) * sign * 1.0 / 1000000
+    num1 = [-83.288056, 42.255085]
+    num2 = [-82.91052, 42.450488]
+    # print "n: " + str(n)
+    num1[0] += n
+    num1[1] += n
+    num2[0] += n
+    num2[1] += n
+    middleText = "" + str(num1[0]) + "," + str(num1[1]) + " " + str(num2[0]) + "," + str(num2[1])
+    # print "middleText: " + middleText
+    print text1 + middleText + text2
+


HotLemonJuice:
I have written a python script to generate test 200 test data based on one of the tweet data.
The format looks like this:
`{ "create_at": datetime("2015-11-23T16:14:03.000Z"), "id": 668945640186101761, "text": "Just posted a photo @ Campus Martius Park https://t.co/5Ax4E2CdWZ", "in_reply_to_status": -1, "in_reply_to_user": -1, "favorite_count": 0, "coordinate": point("-83.04647491,42.33170228"), "retweet_count": 0, "lang": "en", "is_retweet": false, "user": { "id": 48121888, "name": "Kevin McKague", "screen_name": "KevinOfMI", "lang": "en", "location": "Davison, Michigan", "create_at": date("2009-06-17"), "description": "I need to ride my bike until my brain shuts up and my muscles are screaming. \nRight after these donuts. Dad of 3.\n Visit my blog 18 Wheels and a 12-Speed Bike.", "followers_count": 1178, "friends_count": 1780, "statues_count": 22263 }, "place": { "country": "United States", "country_code": "United States", "full_name": "Detroit, MI", "id": "b463d3bd6064861b", "name": "Detroit", "place_type": "city", "bounding_box": rectangle("-82.592458,42.950683 -82.214922,43.146086") }, "geo_tag": { "stateID": 26, "stateName": "Michigan", "countyID": 26163, "countyName": "Wayne", "cityID": 2622000, "cityName": "Detroit" } }`

The only difference among all these 200 test data is the value `rectangle("-82.592458,42.950683 -82.214922,43.146086")` where all four numbers are incremented by a random number in the range of [-1, 1].

JavierJia:
cool. a good start point!

HotLemonJuice:
I have added these test data into the sample.adm and gzip it into a new sample.adm.gz. And I ran `./script/ingestTwitterToLocalCluster.sh` afterwards. Is this the correct way to ingest the data?

JavierJia:
Yes. that should work. Why just 200 points? you could make more, right?

JavierJia:
will it be better to give each record a `point` by randomize on its own `place. bounding_box.[0]` ?

HotLemonJuice:
> will it be better to give each record a point by randomize on its own place. bounding_box.[0] ?

@JavierJia Do you mean go through each record from like 4 millions of records to give each of them a randomized point value?

JavierJia:
yes. it might be helpful for you to test.

HotLemonJuice:
And do you mean to insert a new `point` attribute under `place.bouding_box` using `place.bouding_box.rectangle[0]`?

JavierJia:
nono, also no for the previous answer. I take it back. sorry 😢 

The no-precise-location problem will still exist because we are not gonna re-ingest the huge data again. 

A more appropriate solution could be as following: 

If the record has the `coordinate` field, e.g. `"coordinate": point("-83.04647491,42.33170228")`, that is the exact point provided by Twitter which we can directly show on the map. 

If it doesn't have one, we can use it's `place. bounding_box` information to create one. It's better to randomize it around the center instead of the `place. bounding_box[0]`. 

If it doesn't even have the `place.bounding_box`, which could be possible, we just  don't show it.

HotLemonJuice:
So for each existing record, we will use `point` if it exists or simulate a `point` based on the center of its `bounding_box` if the `point` attribute doesn't exist, right?
So then we don't need to generate test data any more right?

JavierJia:
correct 😀

chenlica:
👍  My understanding is that we first show points at the city level, as a good starting point.

HotLemonJuice:
I'm now working on displaying dots using some test point data on the map. I'm looking into the leaflet framework to make it work. 

I see that now the front-end only receives count results from the middleware. My assumption is that in the future the front-end will receive an array of point data.

JavierJia:
yes, sure. by reading our [documentation](http://cloudberry.ics.uci.edu/documentation/), can you work out the required JSON request to the cloudberry?

HotLemonJuice:
```
Data Schema: 
{
  "name":"twitter.ds_tweet",
  "schema":{"typeName":"twitter.typeTweet",
    "dimension":[
      {"name":"create_at","isOptional":false,"datatype":"Time"},
      {"name":"id","isOptional":false,"datatype":"Number"},
      {"name":"coordinate","isOptional":false,"datatype":"Point"},
      {"name":"lang","isOptional":false,"datatype":"String"},
      {"name":"hashtags","isOptional":true,"datatype":"Bag","innerType":"String"},
      {"name":"geo_tag.stateID","isOptional":false,"datatype":"Number"},
      {"name":"geo_tag.countyID","isOptional":false,"datatype":"Number"},
      {"name":"geo_tag.cityID","isOptional":false,"datatype":"Number"},
      {"name":"geo","isOptional":false,"datatype":"Hierarchy","innerType":"Number",
        "levels":[
          {"level":"state","field":"geo_tag.stateID"},
          {"level":"county","field":"geo_tag.countyID"},
          {"level":"city","field":"geo_tag.cityID"}]},
      {"name": "place.bounding_box", "isOptional":false,"datatype":"rectangle"}  //my change - But I notice the middleware doesn't have this rectangle datatype. Do we have to create this new datatype?
    ],
    "measurement":[
      {"name":"text","isOptional":false,"datatype":"Text"},
      {"name":"favorite_count","isOptional":false,"datatype":"Number"},
      {"name":"retweet_count","isOptional":false,"datatype":"Number"}
    ],
    "primaryKey":["id"],
    "timeField":"create_at"
  }
}

Query:
{
  "dataset": "twitter.ds_tweet",
  "filter": [
    {
      "field": "create_at",
      "relation": "inRange",
      "values": [ "2016-01-01T00:00:00.000Z", "2016-12-31T00:00:00.000Z"]
    },
    {
      "field": "text",
      "relation": "contains",
      "values": [ "zika", "virus" ]
    }
  ],
  "group": {
     "by": [
        {
          "field": "coordinate",  //my change
          "as": "point"
        },
        {
            "field": "place.bounding_box",   //my change
            "as": "rectangle"
        },
        {
          "field": "create_at",
          "apply": {
            "name": "interval",
            "args": {
              "unit": "hour"
            }
          },
          "as": "hour"
        }
      ]    //my change - I deleted aggregation here as it doesn't make sense to aggregate the count on each point
  }
}
```

JavierJia:
As for `rectangle`, can we express it as a bag of `point`s? 
For the query part, we definitely don't need a `groupby`. Can you take a look at the current js code to see how the `sample` request looks like? (The `sample` request gets the actual tweet id and show it on the sidebar).

HotLemonJuice:
I see. I think we can use the sample query to retrieve those point data. I added `"coordinate", "place.bounding_box"` in the `select.field`.
```
{
    "dataset":"twitter.ds_tweet",
    "filter":[
        {
            "field":"geo_tag.stateID",
            "relation":"in",
            "values":[37,51,24,11,10,34,42,9,44,48,35,4,40,6,20,32,8,49,12,22,28,1,13,45,5,47,21,29,54,17,18,39,19,55,26,27,31,56,41,46,16,30,53,38,25,36,50,33,23,2]
        },
        {
            "field":"create_at",
            "relation":"inRange",
            "values":["2017-03-14T18:57:07.994Z","2017-03-15T18:57:07.994Z"]
        },
        {
            "field":"text",
            "relation":"contains",
            "values":["bike"]
        }
    ],
    "select":{
        "order":["-create_at"],
        "limit":10,
        "offset":0,
        "field":["create_at","id","user.id", "coordinate", "place.bounding_box"] //myChange
    },
    "transform":{
        "wrap":{"key":"sample"}
    }
}
```

For the rectangle, it is essentially a bag of points, but can the middleware recognize it without some manual parsing since its format is like 
```
"bounding_box": rectangle("-124.482003,32.528832 -114.131212,42.009519")
``` 
but not 
```
"bounding_box": {
    point("-124.482003,32.528832"),
    point("-114.131212,42.009519")
}
```
?

JavierJia:
hmm, it indeed a problem. the Rectangle is a special object we need to fully support in Cloudberry. Right now it's not implemented yet. But somehow it makes sense only in AsterixDB context and there is ongoing work to make it compatible with the `GeoJSON` format. We need to come up with a general enough "box" object. 

It will take time. Meanwhile, I think we have to reformat the data field of that `bounding_box` as an array of points?  e.g. [ point(133.0, 33), point(233,33] . 😭

HotLemonJuice:
we can use city ID and coordinate to show the points first. No need to change the data for now.

PR #277, opened 2017-03-08
[Hotfix] Prevent Timeline from Overlapping the Sidebar

HotLemonJuice:
Purpose:
- Prevent the timeline from overlapping with the sidebar

Implementation details:
- Reduce timeline width 

Screenshots:
# 720p
screen shot 2017-03-07 at 20 56 11
screen shot 2017-03-07 at 20 56 19
# Tablet resolution
screen shot 2017-03-07 at 20 59 11

The timeline doesn't overlap with the sidebar under this small resolution but other elements on the page appear to be out of place. If we plan to make the Tweeter Demo compatible with small screen sizes like tablets and phones in the future, we will need to fix these out-of-place elements.


JavierJia:
Looks good! 👍 
Let's solve the small screen devices problem in a lower priority.

PR #290, opened 2017-04-03
Update DataSetInfo to fix MetaSchema

luochen01:
The MetaSchema was wrong in that the timeField ("stats.createTime") was not added into its dimension fields. Now it's fixed.


zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/DataSetInfo.scala
```diff
@@ -24,7 +24,8 @@ object DataSetInfo {
 
   val MetaDataDBName: String = "berry.meta"
   val MetaSchema: Schema = Schema("berry.MetaType",
-    Seq(StringField("name")),
+    Seq(StringField("name"),
```
JavierJia: Can we add a test case for this to prevent the future errors?

luochen01: Sure, added a test case for AQLGenerator.

Issue #268, opened 2017-03-03
Refactor Cloudberry Query-related code to use typed fields


luochen01:
Current, the `Query` class uses field names (with string type) to refer fields. For example, the definition of `FilterStatement` is
```
case class FilterStatement(field: String,
                           funcOpt: Option[TransformFunc],
                           relation: Relation,
                           values: Seq[Any]) extends Statement {
```

This imposes several problems on the subsequent use of the field, since one has to manually lookup from the schema to get the typed field, which could be tedious.

Thus, it's better to refactor query-related code to use typed fields. The expected main changes as follows:

- In the query-related classes, we need to replace the string field with a typed field (with `Field` type). This could result in a lot of changes in the codebase which uses the query-related objects.
- `JsonParser` should resolve the fields from string field names, and validate the user-submitted query (from json objects), and return a query which contains typed fields.

JavierJia:
Can we keep the `JSONParser` and let the resolving and validating be the internal functions of it? 
I think the `QueryResolver` and `Validator` should only be used once inside the parser.

luochen01:
Yes, `QueryResolver` and `Validator` will be inside `JsonParser`. My initial thought is that it would be cleaner for `JsonParser` if we create the `UnresolveQuery` first and then resolve it using `QueryResolver`, rather than mixing up everything (json parsing, field resolving and query validating) within the same `JsonParser`.

chenlica:
Good job!

PR #292, opened 2017-04-04
Fix a bug that only the first query in a batch is executed by BerryClient

luochen01:
For a query batch with N queries, query slicing should send N mini queries at each time slot. However, after issue #268, we experienced a bug that only the first mini query was sent to the backend. The reason is that previously there is an implicit assumption that the number of datasetInfos resolved should be the same as the number of queries. In other words, each query should correspond to one datasetInfo (which may cause the same datasetInfo fetched multiple times). This implicit assumption was broken after issue #268 by avoiding multiple fetches of the same datasetInfo.

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
@@ -39,7 +39,7 @@ class BerryClient(val jsonParser: JSONParser,
 
   private case class QueryGroup(ts: DateTime, curSender: ActorRef, queries: Seq[QueryInfo], postTransform: IPostTransform)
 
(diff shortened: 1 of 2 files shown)


zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -52,12 +52,12 @@ class BerryClient(val jsonParser: JSONParser,
     case request: Request =>
       handleNewRequest(request, out.getOrElse(sender()))
     case initial: Initial if initial.ts == curKey =>
-      val queryInfos = initial.queries.zip(initial.infos).map {
```
JavierJia: I see it now, the `zip` caused the `3` to `1` issue.

JavierJia:
Can you add some description for the PR?

PR #298, opened 2017-04-06
[issue#295] Implement the Lookup inside Groupby

luochen01:
This PR implemented the lookup statements inside GroupBy. The semantics is that after performing group by, we then perform lookup operations (left outer join) on the grouped results. Thus, now we have two kinds of lookup operations, one before groupby (i.e., lookup for each record) and another after group by (i.e., lookup for each group).

The main changes are as follows:
1. Modified `Query` to add lookup statements inside groups.
2. Modified `SQLPPGenerator` to generate code for lookups inside groups.
3. Refactored `SQLPPGenerator` to incrementally build a query (previously we generate query parts, which are composed together in the end). This scheme is more suitable for generating nested queries.
4. Modified `JSONParser` to support the syntax changes.
5. Added corresponding test cases `SQLPPGenerator` and `JSONParser` regarding this change.

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
@@ -142,7 +142,7 @@ class SQLPPGenerator extends AsterixQueryGenerator {
         }
         s"""left outer join ${lookup.dataset} $lookupExpr on ${conditions.mkString(" and ")}"""
     }.mkString("\n")
-    append(queryBuilder, lookupStr)
+    appendIfNotEmpty(queryBuilder, lookupStr)
 
     ParsedResult(Seq.empty, (producedExprs).result().toMap)
   }
@@ -155,7 +155,7 @@ class SQLPPGenerator extends AsterixQueryGenerator {
         parseFilterRelation(filter, exprMap(filter.field.name).refExpr)
       }
       val filterStr = (unnestTestStrs ++ filterStrs).mkString("where ", " and ", "")
-      append(queryBuilder, filterStr)
+      appendIfNotEmpty(queryBuilder, filterStr)
 
       ParsedResult(Seq.empty, exprMap)
     }
@@ -175,7 +175,7 @@ class SQLPPGenerator extends AsterixQueryGenerator {
         }
         s"unnest ${expr.refExpr} $newExpr"
     }.mkString("\n")
-    append(queryBuilder, unnestStr)
+    appendIfNotEmpty(queryBuilder, unnestStr)
 
     ParsedResult(unnestTestStrs.toSeq, (producedExprs ++= exprMap).result().toMap)
   }
@@ -195,7 +195,7 @@ class SQLPPGenerator extends AsterixQueryGenerator {
           s"$groupExpr as $newExpr"
         }
         val groupStr = s"group by ${groupStrs.mkString(",")} group as $groupVar"
-        append(queryBuilder, groupStr)
+        appendIfNotEmpty(queryBuilder, groupStr)
 
         group.aggregates.foreach { aggr =>
           val fieldExpr = exprMap(aggr.field.name)
@@ -247,9 +247,9 @@ class SQLPPGenerator extends AsterixQueryGenerator {
 
         val limitStr = s"limit ${select.limit}"
         val offsetStr = s"offset ${select.offset}"
-        append(queryBuilder, orderStr)
-        append(queryBuilder, limitStr)
-        append(queryBuilder, offsetStr)
+        appendIfNotEmpty(queryBuilder, orderStr)
+        appendIfNotEmpty(queryBuilder, limitStr)
+        appendIfNotEmpty(queryBuilder, offsetStr)
 
         if (select.fields.isEmpty) {
           producedExprs ++= exprMap
@@ -355,7 +355,7 @@ class SQLPPGenerator extends AsterixQueryGenerator {
     }
   }
 
-  protected def append(queryBuilder: StringBuilder, queryStr: String): Unit = {
+  protected def appendIfNotEmpty(queryBuilder: StringBuilder, queryStr: String): Unit = {
     if (!queryStr.isEmpty) {
       queryBuilder.append("\n")
       queryBuilder.append(queryStr)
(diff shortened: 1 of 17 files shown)


zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
```diff
@@ -264,7 +279,8 @@ class SQLPPGenerator extends AsterixQueryGenerator {
           } else {
             s"select value $sourceVar"
           }
-        ParsedResult(Seq(projectStr, "", "", ""), exprMap)
+        queryBuilder.insert(0, projectStr + "\n")
```
JavierJia: what's the purpose of this `insert`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
```diff
@@ -334,6 +354,13 @@ class SQLPPGenerator extends AsterixQueryGenerator {
       case _ => aggFuncExpr(typeImpl.getAggregateStr(aggr.func))
     }
   }
+
+  protected def append(queryBuilder: StringBuilder, queryStr: String): Unit = {
+    if (!queryStr.isEmpty) {
```
JavierJia: why it won't do anything if `isEmpty`?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
```diff
@@ -541,8 +545,116 @@ class SQLPPGeneratorTest extends Specification {
...
+        """
+          |select tt.`state` as `state`,tt.`count` as `count`,ll0.`population` as `population`
+          |from (
+          |select `state` as `state`,coll_count(g) as `count`
+          |from twitter.ds_tweet t
+          |where similarity_jaccard(word_tokens(t.`text`), word_tokens('zika')) > 0.0
+          |and contains(t.`text`, "virus")
+          |group by t.geo_tag.stateID as `state` group as g
+          |) tt
+          |left outer join twitter.US_population ll0 on ll0.`stateID` = tt.`state`;""".stripMargin.trim
```
JavierJia: nice!

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
```diff
@@ -541,8 +545,116 @@ class SQLPPGeneratorTest extends Specification {
...
+          |from twitter.ds_tweet t
+          |left outer join twitter.US_population l0 on l0.`stateID` = t.`geo_tag`.`stateID`
+          |where similarity_jaccard(word_tokens(t.`text`), word_tokens('zika')) > 0.0
+          |and contains(t.`text`, "virus")
+          |group by t.geo_tag.stateID as `state` group as g
+          |) tt
+          |left outer join twitter.US_literacy ll0 on ll0.`stateID` = tt.`state`
+          |limit 0
+          |offset 0;""".stripMargin.trim
+      )
```
JavierJia: so cool!

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
luochen01: Here queryBuilder is used to incrementally build the query. `projectStr` is actually a select statement, and thus has to be appeared in the beginning of the query.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
luochen01: If `queryStr` is empty, it means the corresponding statement is empty as well. For example, an empty filter:Seq[FilterStatement] or unnest:Seq[UnnestStatement].

JavierJia: then let's make a good name like `appendIfQueryIsNotEmpty` or something else that describe the behavior?

JavierJia: And add a comment to this function explain why we don't need to append. IMO it's more like a trick to write less code. :-)

luochen01: Make sense. I used the name `appendIfNotEmpty` to describe this behavior.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
```diff
@@ -355,6 +355,11 @@ class SQLPPGenerator extends AsterixQueryGenerator {
     }
   }
 
+  /**
+    * Append a new line and queryStr to the queryBuilder if queryStr is not empty
```
JavierJia: we don't need to repeat the function name, right? 
it's better to specify the behind-scene reason  about why we don't append if the querystring is empty?

luochen01: Oh I see, just added one more explanation on this.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -738,6 +755,178 @@ object TestQuery {
      """.stripMargin
   )
 
+  val groupLookupJSON = Json.parse(
```
JavierJia: @HotLemonJuice Here is the example of using Lookup inside the groupby. 😄

JavierJia:
@HotLemonJuice you should take a look at the JSON request exmaple about `groupbyLookup` in the test case and write the similar query in the front-end.  Ask question to @luochen01 if you have any. 😄

chenlica:
Super!  I would like to have a F2F discussion about how you implemented it.  Please ping me to discuss.  Great job!

Issue #300, opened 2017-04-06
Update the text index type to full-text index


JavierJia:
Previously, we were using the `text` index. As @waans11  implemented the full-text search feature in the newest AsterixDB, we need to build the full-text index and update the AQLGenerator and the SQLPPGenerator to accommodate the new searching syntax.

JavierJia:
The code is here: `zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala` `parseTextRelation`.

JavierJia:
To use the `full-text` search, we will need to create a full-text index. In the xxx
 we will need to update this line 
`create index text_idx if not exists on ds_tweet("text") type keyword;`
to
`create index text_idx if not exists on ds_tweet("text") type fulltext;`

waans11:
The documentation can be found at here:
https://ci.apache.org/projects/asterixdb/aql/fulltext.html

waans11:
So, we need to remove jaccard > 0 part and replace contains to ftcontains.

chenlica:
Great to see the full-text search in action :-)

PR #302, opened 2017-04-07
Use fulltext index in AsterixDB

JavierJia:
This PR replace the `keyword` index with the newest `fulltext` index supported in the recent AsterixDB.

The index type in the ingestion script is updated. And the `parseTextRelation` is converted to use `ftcontains` function.

#300 

Blocked by the issue [AsterixDB-1877](https://issues.apache.org/jira/browse/ASTERIXDB-1877)


luochen01:
Looks good to me. Just one question, why remove `-set wait-for-completion-feed "false";`? It could be very slow to wait for feed ingestion, even it's sample feed...

JavierJia:
Base on @idleft 's explanation, that line is unnecessary. The socket feed will keep open by default.

JavierJia:
Will wait for the AsterixDB fix and then merge :-)

JavierJia:
will use a [fix](https://issues.apache.org/jira/browse/ASTERIXDB-1877) to set up the service.

chenlica:
A big milestone!  Wish the new feature is more powerful and stable :-)

PR #304, opened 2017-04-09
Added an option to switch between AQL/SQLPP

luochen01:
Added an option in application.conf to switch between AQL and SQLPP.

Usage:
asterixdb.lang = "SQLPP"
or
asterixdb.lang = "AQL"

Note that the corresponding asterixdb.url also needs to be changed.


neo/app/controllers/Application.scala
```diff
@@ -34,11 +34,16 @@ class Application @Inject()(val wsClient: WSClient,
 
   val config = new Config(configuration)
   val cities = Application.loadCity(environment.getFile(config.USCityDataPath))
-  val asterixConn = new AsterixConn(config.AsterixURL, wsClient)
+  val (asterixConn, qlGenerator) =
+    config.AsterixLang match {
```
JavierJia: add one `tolowecase` tobe safe?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IDataConn.scala
```diff
@@ -19,9 +19,50 @@ trait IDataConn {
   def post(statement: String): Future[WSResponse]
 }
 
-class AsterixConn(url: String, wSClient: WSClient)(implicit ec: ExecutionContext) extends IDataConn with Logging {
+class AsterixAQLConn(url: String, wSClient: WSClient)(implicit ec: ExecutionContext) extends IDataConn with Logging {
```
JavierJia: this file is a little big and will grow bigger if we have more connections. 
I think we can have a `AsterixConnection` file and put `AQLConn` and `SQLPPConn` there.

neo/app/controllers/Application.scala
luochen01: Yes, in case of mis-spelling ("aql" and "AQL" are both acceptable)

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IDataConn.scala
luochen01: Sure!

PR #306, opened 2017-04-13
Add register process in launch instructions

haochen07:
Tell users how to register twitter map data model when they launch cloudberry at first time.


JavierJia:
👍

haochen07:
I cannot merge since one of the checks is not successful...

PR #308, opened 2017-04-14
[#ISSUE 305]Add "DeleteRecord" query for deleting records from the backend db

luochen01:
Added a "DeleteRecord" for deleting records from the backup db.

This query has two parameters, i.e., dataset and a seq of filters. The semantics is that records satisfying the filters condition would be deleted from the given data set (basically the same way of SQL)

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
@@ -97,10 +97,18 @@ class SQLPPGenerator extends AsterixQueryGenerator {
        |)""".stripMargin
   }
 
+  protected def parseDelete(delete: DeleteRecord, schemaMap: Map[String, Schema]): String = {
+    val exprMap: Map[String, FieldExpr] = initExprMap(delete.dataset, schemaMap)
+    val queryBuilder = new StringBuilder()
+    queryBuilder.append(s"delete from ${delete.dataset} $sourceVar")
+    parseFilter(delete.filters, exprMap, Seq.empty, queryBuilder)
+    return queryBuilder.toString()
+  }
+
   def parseQuery(query: Query, schemaMap: Map[String, Schema]): String = {
     val queryBuilder = new mutable.StringBuilder()
 
-    val exprMap: Map[String, FieldExpr] = initExprMap(query, schemaMap)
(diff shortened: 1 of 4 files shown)


zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
```diff
@@ -695,4 +695,24 @@ class SQLPPGeneratorTest extends Specification {
         """.stripMargin.trim)
     }
   }
+
+  "SQLPPGenerator deleteRecord" should {
+    "generate the delete query without filter" in {
+      val sql = parser.generate(DeleteRecord(TwitterDataSet, Seq.empty), Map(TwitterDataSet -> TwitterDataStore.TwitterSchema))
+      removeEmptyLine(sql) must_== unifyNewLine(
+        """
+          |delete from twitter.ds_tweet t;
```
JavierJia: is this statement valid?

luochen01: Yes, it's executable by AsterixDB (simply delete everything from the dataset)

JavierJia: 😨 it's a scary request. could we ban this kind of without-filter delete request?

luochen01: Sure, but it should be fine. This query is not exposed to public, but for internal use only...

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
```diff
@@ -97,10 +97,21 @@ class SQLPPGenerator extends AsterixQueryGenerator {
        |)""".stripMargin
   }
 
+  protected def parseDelete(delete: DeleteRecord, schemaMap: Map[String, Schema]): String = {
+    if (delete.filters.isEmpty) {
+      throw new QueryParsingException("Filter condition is required for DeleteRecord query.")
```
JavierJia: 👍

haochen07:
Looks good to me!

PR #310, opened 2017-04-14
Add test case to retrieve registered data schema

haochen07:


JavierJia:
To reduce the dependency on me, I will let @luochen01 review this PR :-)

haochen07:
Sure.

haochen07:
@luochen01 any comments on that?

PR #311, opened 2017-04-15
Update front end requests info in docs

haochen07:
Add instruction to "transform" field of front end json requests in online documentation.

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -320,7 +320,8 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
 ```
 
 #### Transform response format
-Very often, we want our response in a designated format so that the front end can differentiate responses of different queries easily. Cloudberry support this response transformation by adding an optional "transform" field to json request, whose purpose is to define the transformation format of response. The main transform function we support now is "wrap", which wrappes the whole response as a key-value pair in json with a pre-defined key. For example, if the json request (to get 100 latest sample tweets that mention "zika") is defined in the following way:
+The front end can **optionally** add a "transform" operation in JSON request to define the post-processing operations. 
+For example, front-ends can define a `wrap` operation to wrap the whole response in a key-value pair JSON object in which the `key` is pre-defined. The following request asks the Cloudberry to wrap the result in the value with the key of `sample`:
 
 ```
 {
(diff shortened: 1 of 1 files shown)


docs/documentation.md
```diff
@@ -319,6 +319,44 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
 }
 ```
 
+#### Transform response format
+Very often, we want our response in a designated format so that the front end can differentiate responses of different queries easily. Cloudberry support this response transformation by adding an optional "transform" field to json request, whose purpose is to define the transformation format of response. The main transform function we support now is "wrap", which wrappes the whole response as a key-value pair in json with a pre-defined key. For example, if the json request (to get 100 latest sample tweets that mention "zika") is defined in the following way:
```
JavierJia:
```
The front end can **optionally** add a "transform" operation in JSON request to define the post-processing operations. 

For example, front-ends can define a `wrap` operation to wrap the whole response in a key-value pair JSON object in which the `key` is pre-defined. The following request asks the Cloudberry to wrap the result in the value with the key of `sample`:
```

docs/documentation.md
```diff
@@ -319,6 +319,44 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
...
+  }
+  "transform" : {
+    "warp": {
+      "key": "sample"
+    }
+  }
+}
+```
+
+Then the response is in the format below. It is now very easy to identify that this is the response of the request above.
```
JavierJia: Too many words :-)

The response is as below:

docs/documentation.md
```diff
@@ -319,6 +319,44 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
...
+
+```
+{
+  "sample":{ 
+    response of samples
```
JavierJia: can you list a real example of the `response of samples` ?

docs/documentation.md
```diff
@@ -319,6 +319,44 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
...
+  }
+}
+```
+
+We encourage you to take advantage of the "transform" field because in this way, you can add your customized header to your response and make it efficient to correspond them with requests.
+Note that "transform" field is optional. If no "transform" field designated, the response will remain its raw format and return to front end.
```
JavierJia: remove this field? since we said so in the previous section.

docs/documentation.md
```diff
@@ -319,6 +319,44 @@ Sometimes the front-end wants to slice multiple queries simultaneously so that i
...
+
+```
+{
+  "sample":{ 
+    response of samples
+  }
+}
+```
+
+We encourage you to take advantage of the "transform" field because in this way, you can add your customized header to your response and make it efficient to correspond them with requests.
```
JavierJia:
```
`wrap` transformation is often preferable when the front-end send many different requests in the same WebSocket interface. 
```

Issue #156, opened 2016-08-08
Add an option to show the normalized count


JavierJia:
... by the number of users or by population. 

It could be implemented by using the `lookup` query.

JavierJia:
The first step can be done by purely adding a frontend option box to normalize the geographic count result. 

shengjix@uci.edu

JavierJia:
@HotLemonJuice

HotLemonJuice:
I'm writing the crawler to extract the population data from [https://www.citypopulation.de/USA-Cities.html](url)
here is the state population data I crawled. It's in json format.
```
[
{"state_population": "4,863,300", "state_name": "Alabama", "state_abbr": "AL"},
{"state_population": "741,894", "state_name": "Alaska", "state_abbr": "AK"},
{"state_population": "6,931,071", "state_name": "Arizona", "state_abbr": "AZ"},
{"state_population": "2,988,248", "state_name": "Arkansas", "state_abbr": "AR"},
......
]
```
The population data is the latest one the website can provide.
@JavierJia How does it look?

JavierJia:
pretty nice!
can you also crawl the city populations?

JavierJia:
I added @luochen01 , it is the use case of new `lookup` statement. #295 
And @luochen01 can also help Shengjie to store the population json into the AsterixDB. You two can have a close communication :-)

HotLemonJuice:
OK, now I have crawled city and county population as well. 
## city population
```
[
{"city_name": "Abilene", "city_population": "121,721", "city_admin": "TX"},
{"city_name": "Akron", "city_population": "197,542", "city_admin": "OH"},
{"city_name": "Albuquerque", "city_population": "559,121", "city_admin": "NM"},
...
]
```
## county population
```
[
{"county_population": "55,416", "county_admin": "Alabama", "county_name": "Autauga"},
{"county_population": "208,563", "county_admin": "Alabama", "county_name": "Baldwin"},
{"county_population": "25,965", "county_admin": "Alabama", "county_name": "Barbour"},
{"county_population": "22,643", "county_admin": "Alabama", "county_name": "Bibb"},
{"county_population": "57,704", "county_admin": "Alabama", "county_name": "Blount"},
...
]
```
The state and county population data are complete.
But the city population is:
> The population of all cities, towns and unincorporated places in the United States of America with more than 100,000 inhabitants according to census results and latest official estimates.

[...]

PR #312, opened 2017-04-17
Normalization + Population scripts and data

HotLemonJuice:
Related issue: #156 

Objectives: 
- Show normalized count (i.e. count / population) using lookup query

Changes made:
- `map/controller.js` for UI and `common/service.js` for lookup
- Included state/county/city population data, crawler, and data-cleaning scripts in `noah/src/main/resources/population`

Implementation detail:
- Use Bootstrap Toggle framework to add the toggle in the front end
- Use lookup inside `byGeoRequest` in `service.js` for population data; depends on the `geolevel`, it will query a different dataset from `dsStatePopulation`/`dsCountyPopulation`/`dsCityPopulation`
- Redraw the map whenever the normalization toggle is toggled; by default, the normalization is turned off
- Crawl city/county data using Scrapy (a Python framework) from https://www.citypopulation.de/php/usa-census-alabama.php and state data from https://www.citypopulation.de/USA-Cities.html 
- Clean the data (around 30K cities, 3K counties, and 52states) to make every record in `neo/public/data/*.json` has a matching population data. Side note: Use the average population from counties for 880 cities and the average population from states for 76 counties to deal with the special cases. 

Setup procedure: 
- Ingest data into AsterixDB; data is in `noah/src/main/resources/population/adm`; example ingestion queries are in `noah/src/main/resources/population/sqlpp`
- register `dsStatePopulation`/`dsCountyPopulation`/`dsCityPopulation` using scripts in `scripts/register*.sh`

nmlz1 mov
nmlz2 mov
nmlz3 mov
screen shot 2017-04-17 at 13 40 39
screen shot 2017-04-17 at 13 40 51


neo/app/views/main.scala.html
```diff
@@ -17,6 +17,7 @@
   <link rel="stylesheet" media="screen" href='@routes.Assets.versioned("lib/font-awesome/css/font-awesome.min.css")'>
   <link rel="stylesheet" media="screen" href='@routes.Assets.versioned("lib/bootstrap-vertical-tabs/bootstrap.vertical-tabs.min.css")'>
   <link rel='stylesheet' href='@routes.Assets.versioned("stylesheets/main.css")'>
+  <link href="https://gitcdn.github.io/bootstrap-toggle/2.2.2/css/bootstrap-toggle.min.css" rel="stylesheet">
```
HotLemonJuice: I think I may need to use the Scala way to import library...

JavierJia:
Nice :-). Maybe @luochen01 can give a review first?

neo/public/javascripts/map/controllers.js
```diff
@@ -1,6 +1,9 @@
 angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
   .controller('MapCtrl', function($scope, $window, $http, $compile, Asterix, leafletData) {
     $scope.result = {};
+    $scope.doNormalization = false;
+    $scope.upscaleFactor = 1000 * 1000;
+    $scope.upscaleFactorText = "M";
```
luochen01: For scaleFactor and scaleFactorText, I think it's better to have a global config file or class for us to manage the demo more easily?

neo/public/javascripts/map/controllers.js
```diff
@@ -394,56 +399,94 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
+        }
+        else{
+          d.properties.countText = d.properties.count.toFixed(1);
+        }
+        d.properties.countText += "M";
+      }
+
       //FIXME: the code in county and city (and probably the state) levels are quite similar. Find a way to combine them.
-      if ($scope.status.logicLevel == "state" && $scope.geojsonData.state) {
+      if ($scope.status.logicLevel === "state" && $scope.geojsonData.state) {
```
luochen01: I'm not familiar with JS... but why there is  ===?

neo/public/javascripts/common/services.js
```diff
@@ -74,6 +74,18 @@ angular.module('cloudberry.common', [])
       ];
     }
 
+    function getPopulationTarget(parameters){
+      var targetLowerCase = parameters.geoLevel;
+      var targetUpperCase = targetLowerCase.charAt(0).toUpperCase() + targetLowerCase.substr(1);
+      return {
+        joinKey: [targetLowerCase],
+        dataset: "twitter.ds" + targetUpperCase + "Population",
```
luochen01: It's better to have some meta object to map state/county/city to its dataset. Try to avoid string concatenation...

luochen01:
Sure, I made a few comments. However, the problem is I'm not familiar with javascript...

neo/public/javascripts/map/controllers.js
HotLemonJuice: Good point. I wonder if we are going to use Scala or javascript

luochen01: I think it should be in js, since it's front-end related. Do we have some global config file or object for the TwitterMap right now? If the case, we can add these two options to it. If not, maybe we can have one...

neo/public/javascripts/map/controllers.js
HotLemonJuice: === means strict comparison. It's type-safe. In javascript, it's a preferred way to compare two variables when you don't want the type of the variables gives you unexpected results.

JavierJia:
Before jump into the code, in the screen shot, what does `Count:6.6M` mean in the legend?

HotLemonJuice:
I sent the dataset registering requests manually using terminal. I wrote `registerStatePopulation.sh` and two other similar shell scripts for that. Do I need to send the register in the front end?

HotLemonJuice:
Oops. I meaned to say 6.6/M. It means 6.6 tweets per million people. I will change it.

luochen01:
I think maybe we need to come up with automatic way to register those datasets. Otherwise, other devs have to do it over and over again manually.

neo/public/javascripts/common/services.js
HotLemonJuice: Do you mean to explicitly declare three separate query objects for state/county/city and use some conditions to map to them? And I wonder the reason to avoid using string concatenation. Is it because it's more error prone?

neo/public/javascripts/map/controllers.js
HotLemonJuice: A good example taken from https://stackoverflow.com/questions/523643/difference-between-and-in-javascript:
```
0 == false   // true
0 === false  // false, because they are of a different type
1 == "1"     // true, automatic type conversion for value only
1 === "1"    // false, because they are of a different type
null == undefined // true
null === undefined // false
'0' == false // true
'0' === false // false```

File: neo/public/javascripts/common/services.js (hunk 4 above)
Comment (luochen01) [inline neo/public/javascripts/common/services.js]: Not that complex. You may define a map or a meta object to have mappings like state->statePopulation, county->countyPopulation etc. This mapping should also be resided in the global config object (if we have one).

**Comment by HotLemonJuice** [comment]
I have addressed all the issues @luochen01 mentioned. After discussing with @JavierJia , there is no need for the automatic data model registration for now. However, I did combine all the data model registration (ds_twitter and population) into one file: `./script/registerTwitterMapDataModel.sh`. It should make registration a bit easier.

**Comment by luochen01** [comment]
Saw it. And make sure to have clear instructions for those who want to try normalization on their local environment...

File: neo/app/views/main.scala.html (hunk 1 above)
Comment (JavierJia) [inline neo/app/views/main.scala.html]: have you figured this out? @vidhya567 can help on this.

File: neo/public/javascripts/common/services.js [hunk 5]
```diff
@@ -1,5 +1,39 @@
...
+              dataset: "twitter.dsCityPopulation",
+              lookupKey: ["cityID"],
+              select: ["population"],
+              as: ["population"]
+            };
+        }
+      }
+    };
+  })
+  .service('Asterix', function($http, $timeout, $location, cloudberryConfig) {
```
JavierJia: could you take this chance to rename this `Asterix`? it's been a while. It should be `cloudberry`.

neo/public/javascripts/map/controllers.js
```diff
@@ -394,56 +397,94 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
             d.properties.count = 0;
+          if (d.properties.countText)
+            d.properties.countText = "";
           for (var k in result) {
             //TODO make a hash map from ID to make it faster
-            if (result[k].city == d.properties.cityID) {
-              d.properties.count = result[k].count;
+            if (result[k].city === d.properties.cityID) {
+              if($scope.doNormalization){
+                d.properties.count = result[k]['count'] / result[k]['population'] * cloudberryConfig.normalizationUpscaleFactor;
```
JavierJia: could you define a util function like `normalizeCount` to make the three places of the same code look better?

neo/public/javascripts/map/controllers.js
```diff
@@ -464,19 +505,33 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
...
+
         var gName  = grades.map( function(d) {
+          var returnText = "";
           if (d < 1000){
-            return d.toString();
+            returnText = d.toString();
           }
-          if (d < 1000 * 1000) {
-            return (d / 1000).toString() + "K";
+          else if (d < 1000 * 1000) {
```
JavierJia: I think our convention for `if else` block is 
```
if {
} else  if {
} else {
}
```

noah/src/main/resources/population/crawler/population/population/items.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
JavierJia: I can't comment on the empty file, is that `__init__.py` required? 
I'm sure we can remove the `*.pyc`.

noah/src/main/resources/population/crawler/population/population/spiders/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This package will contain the spiders of your Scrapy project
```
JavierJia: remove those `*.pyc`

noah/src/main/resources/population/crawler/population/population/items.py
HotLemonJuice: I have removed `*.pyc`. `__init__.py` is required to make Python treat the directories as containing packages. Find more in this link: https://stackoverflow.com/questions/448271/what-is-init-py-for

HotLemonJuice:
@JavierJia I have addressed all the issues you mentioned. And I have updated the documentation for instructions of setting up the environment for normalization.

chenlica:
Super!  Can we push it to the production system?

JavierJia:
Yes, we can. I will push it soon.

PR #316, opened 2017-04-21
Implement DropView query in SQLPPGenerator

luochen01:
Implemented DropView query in SQLPPGenerator. This query is used to physically remove a view from AsterixDB.

A sample query looks like:
`drop dataset test if exists;`


JavierJia:
I think @haochen07 can review this PR :-)

haochen07:
Looks good to me!

chenlica:
Just curious: how can we prevent external users from making such kind of requests?  Should we do the protection at the network layer>

JavierJia:
It will be a web admin console that requiring some authentications for the purpose of the `register` and `deregister` in the future.

Issue #305, opened 2017-04-13
Send deletion query from cloudberry to AsterixDB to finish deregistering process


haochen07:
Now the DDL is almost finished! 
The cloudberry can receive register/deregister json request from front-end, and in registering process, cloudberry is able to keep that information in `metaData` and flush to database; in deregistering process, cloudberry can remove the dataset from `metaData` but not able to update the information to database because the deletion query is not defined inside cloudberry. Once it is done, the whole register/deregister process is finished from front to end.

haochen07:
@luochen01 Could you enable a deletion query in cloudberry? I can handle all other logics concerning deregister process. Thanks

luochen01:
What kind of delete query is required? Only delete one tuple/record, or delete the whole dataset?
It would be good to have a concrete use case/example here.

haochen07:
In `DataStoreManager`, the `metaData` stores all the dataset infos and periodically flush them to database. What we need is when we remove a dataset info from `metaData`, we send a query to database and the removed dataset info will also be removed from dataset.

JavierJia:
delete one record from the meta dataset.

luochen01:
Then I would assume we can add a filter to the delete statement? Those records satisfying this filter would be deleted (to be consistent with most database systems).

JavierJia:
correct!

luochen01:
Now `DeleteRecord` query is available for SQLPPGenerator. You can go ahead with the deregister dataset feature.

PR #320, opened 2017-04-22
[Issue#305] Deregister process can delete record in meta dataset and drop views

haochen07:
This PR is related to #305.

**Problems**
The previous `deregister` process only removes the data schema from cloudberry's in-memory map, and did not touch the meta dataset and views that are stored in the backend database. This PR finishes the `deregister` pipeline. Now a user is able to send a `deregister` request from the front end, and the dataset in the database will be deleted. 

**Implementations**
- The schema in the in-memory map will be removed.
- The schema stored in the `meta` dataset will be removed.
- All the `Views` associated with the dataset will be physically removed and deleted from the `meta` dataset.


zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,28 @@ class DataStoreManager(metaDataset: String,
...
+      val metaRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dropTableName))
+      metaActor ! DeleteRecord(metaDataset, Seq(metaRecordFilter))
+
+      metaData.foreach( dataSchema =>
+        dataSchema._2.createQueryOpt match {
+          case Some(query) =>
+            if(query.dataset == dropTableName){
+              metaActor ! DropView(dataSchema._1)
+              val viewRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dataSchema._1))
+              metaActor ! DeleteRecord(metaDataset, Seq(viewRecordFilter))
```
JavierJia: another leftover thing I just remember is that we should also kill the corresponding `ViewDataAgent` and the `OriginalDataAgent`.

haochen07: OK. What is the logic when `ViewActor` and `OriginActor` are created? From the code, does it mean that one dataset only have one `ViewActor` or `OriginalActor` with the name "data-(dataset name)"? My understand is that when a `Query` or `AppendView` request comes, an actor named "data-(dataset name)" will be called if existed, or created according to if `createQueryOpt` is empty or not. Am I right?

JavierJia: It's created [here](https://github.com/ISG-ICS/cloudberry/blob/master/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala#L183). Yes, one agent is responsible for one specific dataset. You can just get the actor by `context.child("data-" + query.dataset)`. and send it a `PoisonKill` msg.

haochen07:
You have an idea to test that?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
     val dropTableName = dropTable.dataset
 
     if(metaData.contains(dropTableName)){
-      metaData.remove(dropTableName)
+      context.child("data-" + dropTableName) match {
```
JavierJia: how about `context.child("data-" + dropTableName).map( child => child ! PoisonPill)` ?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
...
+        case Some(datasetActor) => datasetActor ! PoisonPill
```
JavierJia: and we should also kill the views

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
...
+      }
 
-      //TODO send query to DELETE dataset in database
+      val metaRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dropTableName))
+      metaActor ! DeleteRecord(metaDataset, Seq(metaRecordFilter))
 
-      sender ! DataManagerResponse(true, "Deregister Finished: dataset " + dropTableName + " has successfully removed.\n")
+      metaData.foreach( dataSchema =>
+        dataSchema._2.createQueryOpt match {
+          case Some(query) =>
```
JavierJia: If we only care about the `Some` case, we don't need to `match` `None` case.
`dataSchema._2.createQueryOpt.map{ query => xxxx }` will be enough.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
     val dropTableName = dropTable.dataset
 
     if(metaData.contains(dropTableName)){
-      metaData.remove(dropTableName)
+      context.child("data-" + dropTableName) match {
+        case Some(datasetActor) => datasetActor ! PoisonPill
+        case None =>
```
JavierJia: w.r.t. to the test, you can create a view actor and a dataset actor, and test if they received `PoisonPill`.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
...
-      //TODO send query to DELETE dataset in database
+      val metaRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dropTableName))
+      metaActor ! DeleteRecord(metaDataset, Seq(metaRecordFilter))
 
-      sender ! DataManagerResponse(true, "Deregister Finished: dataset " + dropTableName + " has successfully removed.\n")
+      metaData.foreach( dataSchema =>
+        dataSchema._2.createQueryOpt match {
+          case Some(query) =>
+            if(query.dataset == dropTableName){
+              metaActor ! DropView(dataSchema._1)
```
JavierJia: we should add test from 175 to 178 by checking if the metaActor received the designed msg.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,33 @@ class DataStoreManager(metaDataset: String,
...
     if(metaData.contains(dropTableName)){
-      metaData.remove(dropTableName)
+      context.child("data-" + dropTableName) match {
+        case Some(datasetActor) => datasetActor ! PoisonPill
+        case None =>
+      }
 
-      //TODO send query to DELETE dataset in database
+      val metaRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dropTableName))
+      metaActor ! DeleteRecord(metaDataset, Seq(metaRecordFilter))
```
JavierJia: to make logic modularized, after 170, we should remove `dropTableName` from `metaData` instead till 184.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
JavierJia: And from 172, it seems can be written as 
```
metaData.filterNot{ case (key, value) => value.createQueryOpt.exists(q => q.dataset == dropTableName) }
   .map { case (key, value) => 
        /// drop msg.
    }
```

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
haochen07: I thought the different way, maybe I am wrong.
The name of the actor, when it is created, depends on the `query.dataset`, which should be the same for all views and the associated dataset. Then in this way, every dataset and all its views will only have one actor.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
haochen07: Got it. I searched this and it seems that if it is a `Unit` function in `map`, `foreach` is better.

haochen07: The `filterNot` will return collection of elements that do not satisfy the condition, thus no drop messages needed for datasets whose `q.dataset == dropTableName`

haochen07:
The last question is in test cases, how to reassign a dataset actor and a view actor to the test system so that expected messages as `PoisonPill` can be caught.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,29 @@ class DataStoreManager(metaDataset: String,
     val dropTableName = dropTable.dataset
 
     if(metaData.contains(dropTableName)){
+
       metaData.remove(dropTableName)
+      context.child("data-" + dropTableName).foreach( child => child ! PoisonPill)
+      val metaRecordFilter = FilterStatement(DataSetInfo.MetaSchema.fieldMap("name"), None, Relation.matches, Seq(dropTableName))
+      metaActor ! DeleteRecord(metaDataset, Seq(metaRecordFilter))
+
+      metaData.filter{ p =>
```
JavierJia: we'd better use `case (name, info) => ` instead of `p` to avoid unamed `_1` and `_2` which has little information.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,29 @@ class DataStoreManager(metaDataset: String,
...
+        p._2.createQueryOpt.exists( query => query.dataset == dropTableName)
+      }.foreach { p =>
```
JavierJia: ditto.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,29 @@ class DataStoreManager(metaDataset: String,
...
+        metaActor ! DropView(p._1)
+        context.child("data-" + p._1).foreach( child => child ! PoisonPill)
+      }
 
-      //TODO send query to DELETE dataset in database
+      // Before retrieve subset of metaData using .filter or .filterNot, etc.,
+      // Use .toMap method to change metaData into immutable map
+      // Otherwise when metaData is clear, no information will be retained.
```
JavierJia: 👍

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -161,12 +161,29 @@ class DataStoreManager(metaDataset: String,
...
+      val newMetaData = metaData.toMap.filterNot{ p =>
+        p._2.createQueryOpt.exists(q => q.dataset == dropTableName)
```
JavierJia: use `case (name, info) `

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
```diff
@@ -166,14 +166,15 @@ class DataStoreManagerTest extends TestkitExample with SpecificationLike with Mo
       newInfo.stats.lastModifyTime.getMillis must be_>=(now.getMillis)
     }
     "update meta info if receive drop request" in {
+      // this test case is for ???
```
JavierJia: if a dataset is dropped, the metadata should not contain the corresponding record.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
```diff
@@ -291,13 +281,80 @@ class DataStoreManagerTest extends TestkitExample with SpecificationLike with Mo
...
+    meta.receiveOne(1 second)
+    meta.reply(initialInfo)
+    sender.send(dataManager, DataStoreManager.AreYouReady)
+    sender.expectMsg(true)
+    sender.send(dataManager, DataStoreManager.AskInfoAndViews(sourceInfo.name))
+    sender.expectMsg(Seq(sourceInfo, zikaHalfYearViewInfo))
+
+    // I tried to name a self-defined actor as the same name of a dataset actor,
+    // but it seems it didn't receive the expected message.
+    val actor1 = system.actorOf(Props(new Actor{
```
JavierJia: I don't know the expected behavior if we create two actors with the same name?

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
```diff
@@ -291,13 +281,80 @@ class DataStoreManagerTest extends TestkitExample with SpecificationLike with Mo
       sender.expectMsg(DataManagerResponse(false, "Register Denied. Field Parsing Error: " + "Time field of " + schemaNotATimeField.typeName + "is not in TimeField format.\n"))
       ok
     }
-    "respond success if deregister an existing data model" in {
+  }
+
+  "Data schema deregistering process" should {
+    val sender = new TestProbe(system)
+    val view = new TestProbe(system)
+    val base = new TestProbe(system)
```
JavierJia: so if we drop base, then this `base` should expect to see a `PoisonPill`.

JavierJia: And the `view` should also receive a `PoisonPill`. that's will be enough.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
haochen07: Then this test case is implemented in deregistering tests.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
haochen07: My question is that since the actors are created when `answerQuery` is called, what we need to do is to send two queries and then drop dataset. In this case `PoisonPill` for `base` and `view` could be caught, right? Rather than loading a dataset and its view from initialization.

JavierJia: correct.

haochen07: I send `Query` and `AppendView`, the messages forwarded to them are received. But when deregistering, these actors are not found through `context.child("name")`.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
```diff
@@ -291,13 +281,80 @@ class DataStoreManagerTest extends TestkitExample with SpecificationLike with Mo
...
+                       dbSchema: Schema,
+                       qLGenerator: IQLGenerator,
+                       conn: IDataConn,
+                       appConfig: Config
+                      )(implicit ec: ExecutionContext): ActorRef = {
+      import AgentType._
+      agentType match {
+        case Meta => meta.ref
+        case Origin => base.ref
+        case View => view.ref
```
JavierJia: Here.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala
JavierJia: because they are not created by that name. you created these actors using `childMaker` which ignore the given parameters.

haochen07:
Now I think everything is ready. You can do the final check @JavierJia .

JavierJia:
Looks good! 👍

PR #324, opened 2017-04-27
Implemented "Append" statement

luochen01:
Now we have the "append" statement to define derived fields based on the base fields in a dataset and UDFs in the underlying database.

A sample query is as follows:
```
{
 "dataset":"twitter.ds_tweet",
 "append": [ {
      "field":"lang",
      "definition":"length(lang)",
      "type":"Number",
      "as":"lang_len"
    }
 ],
 "filter":[
    {
      "field":"lang_len",
      "relation":">=",
      "values":[1]
    }
  ]
...
}
```

Mainly extended `JSONParser` to extend Cloudberry query language, and implemented the translation in `SQLPPGenerator`.

### DIFF
--- a/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
+++ b/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
@@ -13,7 +13,7 @@ class SQLPPGeneratorTest extends Specification {
   "SQLPPGenerator generate" should {
 
     "translate a simple unnest query" in {
-      val query = new Query(TwitterDataSet, Seq.empty, Seq.empty, Seq(unnestHashTag), None, Some(selectTop10))
+      val query = new Query(TwitterDataSet, Seq.empty, Seq.empty, Seq.empty, Seq(unnestHashTag), None, Some(selectTop10))
       val result = parser.generate(query, Map(TwitterDataSet -> twitterSchema))
       removeEmptyLine(result) must_== unifyNewLine(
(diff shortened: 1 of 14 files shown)


chenlica:
Impressive implementation speed!  What if we have a UDF with more than one field? Also can we change `length` to something more like an external function?

luochen01:
The current implementation is very simple, we simply take the `definition` and translate it into SQL++ query. So that definition can be arbitrary string in SQL++ language (it is also fine to have multiple parameters in that string).
BTW, `field` now is not used, and maybe we can use that to do type checking etc.

chenlica:
So the UDF is applied on the entire record?  If so, maybe we can remove the un-used `field`?

luochen01:
It's applied on certain fields. For example, `length(lang)` is applied to the `lang` field to compute its length (similarly, we might use `sentiment_analysis(text)` to get the sentiment score based on the tweet text field). One trick here is that the user is writing SQL++ code directly, and we simply pass this string to SQL++ query.

chenlica:
Understood: `length(lang)` is just the UDF call.  So we are good.

luochen01:
I'll check with Gift this afternoon to see the current syntax of calling sentiment analysis UDF in SQL++. If this PR is enough to handle this UDF, then I'll merge it.

JavierJia:
one possible syntax trick, like we did in `count(*)`, if the UDF (or some other DB function) needs to apply on the record other than a field, it's `field` can be `*`. 
@chenlica I think we can merge this first and have another PR for the sentiment UDF thing if needed?

luochen01:
Just confirmed with Gift, and this syntax works as expected.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGeneratorTest.scala
```diff
@@ -664,6 +664,45 @@ class SQLPPGeneratorTest extends Specification {
           stripMargin.trim
       )
     }
+
```
luochen01: @HotLemonJuice

PR #327, opened 2017-04-30
Implement dynamic legend (issue #156)

HotLemonJuice:
Related issue #156 

This PR adds a dynamic legend in the normalization mode. 
Boundary = (maxCount - minCount) / (# of color intervals)
[0, 1/6]
[1/6, 2/6]
[2/6, 3/6]
[3/6, 4/6]
[4/6, 5/6]
[5/6+]

Side by side comparison: (left: before, right: after)
screen shot 2017-04-30 at 16 08 08
screen shot 2017-04-30 at 16 09 38

### DIFF
--- a/twittermap/public/javascripts/map/controllers.js
+++ b/twittermap/public/javascripts/map/controllers.js
@@ -374,6 +374,10 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
 
       var colors = $scope.styles.colors;
       var sentimentColors = $scope.styles.sentimentColors;
+      var normalizedCountMax = 0,
+          normalizedCountMin = 0,
+          intervals = colors.length - 1,
+          difference = 0;
 
       function getSentimentColor(d) {
         if( d < cloudberryConfig.sentimentUpperBound / 3) {    // 1/3
@@ -385,7 +389,17 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
         }
       }
 
-      function getCountColor(d) {
+      function getNormalizedCountColor(d) {
+        var i = 1;
+        for (; i <= intervals; i++){
+          if ( d <= normalizedCountMin + ((i * difference) / intervals)){  // bound = min + (i / 6) * difference
+            return colors[i];
+          }
+        }
+        return colors[intervals]; // in case of max
+      }
+
+      function getUnnormalizedCountColor(d) {
         if(!d || d <= 0) {
           d = 0;
         } else if (d ===1 ){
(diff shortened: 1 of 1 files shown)


HotLemonJuice:
@chenlica There are some improvements in the city level, but the improvement on the state level is still minimal. The reason is that sometimes the variance of data is very small. In other words, many states have very similar counts, so they end up with the same color. One way we can fix this entirely is to come up with some algorithm to rank the count so that we are sure to divide them equally. But the trade-off of this approach is that the intervals of the legend might be very strange. For example, it can be something like
[0, 1]
[1, 1.5]
[1.5, 1.8]
[1.8, 10]
[10, 100]
[100, 500]
Any thoughts?

chenlica:
@HotLemonJuice I think if the county-level counts are similar, let it be.  The city-level differences look good!

twittermap/public/javascripts/map/controllers.js
```diff
@@ -447,7 +462,10 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
       }
 
       function setNormalizedCount(geo, r){
-        geo['properties']['count'] = r['count'] / r['population'] * cloudberryConfig.normalizationUpscaleFactor;
+        var normalizedCount = r['count'] / r['population'] * cloudberryConfig.normalizationUpscaleFactor;
+        geo['properties']['count'] = normalizedCount;
+        if(normalizedCount > normalizedCountMax)  // update max to enable dynamic legends
+          normalizedCountMax = normalizedCount;
```
JavierJia: do we also need to dynamically change it on `count`?

HotLemonJuice: Are you saying the `'count'` property for each geo? If so, we don't. We only need to keep track of the maximum count, and this is what this line of code is doing.

HotLemonJuice:
So if we are going to use rankings to divide the data, we are sure to have an equal number of polygons in each color interval. But the problem is that it only shows the difference of rankings and it diverges from the message of our original visualization, which was to offer an objective insight into the data.

As Ben Shneiderman (a famous computer scientist in HCI) said:
> The purpose of visualization is insight, not pictures.

I think we shouldn't use ranking, but let the data speak for itself.
@JavierJia @chenlica What do you think?

chenlica:
@HotLemonJuice Seems you have more insights on how to show the data.  Please make a proposal and go ahead to implement it.

JavierJia:
@HotLemonJuice do you have a new proposal or we should proceed to review this change? 😄

HotLemonJuice:
@chenlica I think the current visualization implementation is very good so that's why I said we shouldn't further divide the colors. In other words, let's use the dynamic coloring in this PR. 
 
On the other hand, I have also thought about new visualization implementations. Maybe we can visualize the hashtags because they are topics related to the keyword and they may provide some insight about what people are caring about around this keyword. A good example is a wall of words in which words of higher frequency have larger sizes.

example:
many-eyes

HotLemonJuice:
@JavierJia we should definitely continue reviewing this PR!

JavierJia:
I will try the UDF instruction to add the UDF to AsterixDB docker image so that you can play with the real query. (tonight or tomorrow)

chenlica:
@HotLemonJuice OK, let's use the current way to display colors.  Regarding the new feature, it looks cool.  In general, we want to improve the way to visualize the results without overwhelming the user.  Let's first unify the few features ("count", "normalize", "sentiment") in one sidebar, then think about other improvements.  I may have some other UI tasks for you soon.  I will let you know.

HotLemonJuice:
@JavierJia OK cool!

PR #332, opened 2017-05-04
Remove git lfs

JavierJia:
This PR fixed several issues to make `QuickStart` section work.
1. remove `git lfs` which always caused troubles.
2. add a script to fetch the sample.adm.gz from google drive.
3. fix the feed configuration to open multiples feed ports.


luochen01:
Looks good to me.

Issue #262, opened 2017-02-28
Collect the stats of the DB dataset when registering


JavierJia:
Currently, the original data's stats is a fake one. 
We should collect the `minTime`, `maxTime` and the `count` information when it being registered.

PR #333, opened 2017-05-04
[Issue #262] Collect Stats while register dataset schema

haochen07:
This PR is related to #262 .

**Problem**
When register a new dataset, the `Stats` information in a dataset is now a fake one. We need to query for the correct information and append them to `Stats` field while registering.

**Implementation**
- In `DataStoreManager`
  - Query for `min`, `max` and `cardinality` information while registering.
  - If the `timeField` in schema is not presented in dataset, then use fake information and forward this information to frontend response.
  - When receive new stats sent from `OriginDataAgent`, update info in `meta`.
- In `OriginDataAgent`
  - Change the constructor of `OriginDataAgent`: Pass `DataSetInfo` (mainly for `Interval` and `cardinality`) to `OriginDataAgent` when it starts.
  - In `preStart` of `OriginDataAgent`, only query for cardinality from maxTimeStamp till now.
  - When update `Cardinality`, send message to `dataStoreManager`.


JavierJia:
you don't need to close this issue. I think we can have this capability of storing the stats information in this PR. And address the static table issues afterward.

haochen07:
Reopened. You can have a review on it.

haochen07:
1. For the logics in `originDataAgent`, because now initial query in `preStart` will be removed, the scheduler of `UpdateStats` will not be triggered. We need to start that at the beginning (`preStart`) ? 
2. To transfer message to `DataStoreManager` when `maintenanceWork` is called, I think maybe we need a `actorRef` to  `DataStoreManager` in `OriginDataAgent`

JavierJia:
Good points. 
1. I think we don't need to collect the stats from scratch, but we still want to collect the count from the last maxTime till now, so we don't need to remove the `prestart` logic entirely. then the changes should less.
2. since `DataStoreManager` is the supervisor of all `DatasetAgent`, you can just use `context.parent` to refer it.

haochen07:
Since the registered `Stats` is in `schema`, what we do in the `preStart` is to send this information to `self` and trigger the scheduler? I don't quite understand.

JavierJia:
it could be handled by the `DataStoreManager`. Both ways have pros and cons. 
the principle we want to follow here is that since `DataStoreManager` is THE `manager`, we don't want him to handle too much detailed work. if the logic can be handled by children, just let them do the work and report to the manager whenever necessary. In this way, the code in `manager` will be easier to maintain, though it's already complex now.

haochen07:
Only the testcases of sending new `Stats` to `datastore manager` left. You can take a review of the whole change.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -12,31 +13,18 @@ import scala.concurrent.ExecutionContext
 
 class OriginalDataAgent(override val dbName: String,
                         override val schema: Schema,
+                        val initCardinality: Cardinality,
```
JavierJia: I would suggest just pass a `DataSetInfo` or `Stats` to the agent which makes more sense. And we can put more logic here in the agent about how to use the `info` or `stats`.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -12,31 +13,18 @@ import scala.concurrent.ExecutionContext
...
 
-  val lastCount: Cardinality = UnInitialCardinality
+  val lastCount: Cardinality = initCardinality
 
-  /**
-    * Ask for the initial stats when the Agent starts.
-    * Stats including: minTimeStamp, maxTimeStamp, cardinality
-    */
-  override def preStart(): Unit = {
-    val minTimeQuery = Query(dbName, globalAggr = Some(GlobalAggregateStatement(AggregateStatement(schema.timeField, Min, Field.as(Min(schema.timeField), "min")))))
```
JavierJia: I think we still need the `preStart` to get the delta stats since last update time indicating in the `stats.getMaxTime` or something equivalent.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -58,12 +46,8 @@ class OriginalDataAgent(override val dbName: String,
 
   override protected def maintenanceWork: Receive = {
     case newCount: Cardinality =>
-      if (lastCount.count == UnInitialCount) {
-        lastCount.reset(newCount.from, newCount.till, newCount.count)
-        context.system.scheduler.schedule(config.AgentCollectStatsInterval, config.AgentCollectStatsInterval, self, UpdateStats)
```
JavierJia: we need this `schedule` to periodically fetch the new stats.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgentTest.scala
```diff
@@ -68,91 +64,48 @@ class OriginalDataAgentTest extends Specification with Mockito {
 
       ok
     }
-    "collect stats information when start" in new TestkitExample {
```
JavierJia: it still need to collect the stats, but only the delta part.

haochen07:
code updated as discussed.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -220,14 +220,13 @@ class DataStoreManager(metaDataset: String,
     val actor = context.child("data-" + query.dataset).getOrElse {
       val info = metaData(query.dataset)
       val schema: Schema = info.schema
-      val initCardinality: Cardinality = new Cardinality(info.dataInterval.getStart, info.dataInterval.getEnd, info.stats.cardinality)
       info.createQueryOpt match {
         case Some(_) =>
-          val ret = childMaker(AgentType.View, context, "data-" + query.dataset, query.dataset, schema, initCardinality, queryGenFactory(), conn, config)
+          val ret = childMaker(AgentType.View, context, "data-" + query.dataset, query.dataset, schema, None, queryGenFactory(), conn, config)
```
JavierJia: why this is `None`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -2,29 +2,39 @@ package edu.uci.ics.cloudberry.zion.actor
...
-  context.system.scheduler.schedule(config.AgentCollectStatsInterval, config.AgentCollectStatsInterval, self, UpdateStats)
+  /**
+    * When the Agent starts,
+    *   1. Ask for the stats from maxTimeStamp till now,
+    *   2. Start a scheduler to query for cardinality periodically.
+    * Stats including: minTimeStamp, maxTimeStamp, cardinality.
+    */
+  override def preStart(): Unit = {
+    collectStats(lastCount.till)
+    context.system.scheduler.schedule(config.AgentCollectStatsInterval, config.AgentCollectStatsInterval, self, UpdateStats)
```
JavierJia: I think you can merge this two as 
`context.system.scheduler.schedule(0, config.AgentCollectStatsInterval, self, UpdateStats)`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
haochen07: because the `viewAgent` doesn't need the `dataSetInfo` now. You can see the childMaker function.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
haochen07: Good idea!

PR #340, opened 2017-05-06
Fix a bug of AppendField not projecting fields properly

luochen01:
Previously, when a query has `AppendField` statements, the generated SQL++ query contains a from subquery which projects all fields of the source dataset plus the append fields. E.g.,
```
{
 "dataset":"twitter.ds_tweet",
 "append": [ {
      "field":"lang",
      "definition":"length(lang)",
      "type":"Number",
      "as":"lang_len"
    }
 ],
 "filter":[
    {
      "field":"lang_len",
      "relation":">=",
      "values":[1]
    }
  ]
```
is generated as:
```
select ...
from (select length(lang) as `lang_len`,t.`favorite_count` as `favorite_count`,t.`geo_tag`.`countyID` as `geo_tag.countyID`,t.`user_mentions` as `user_mentions`,t as `geo`,t.`user`.`id` as `user.id`,t.`geo_tag`.`cityID` as `geo_tag.cityID`,t.`is_retweet` as `is_retweet`,t.`text` as `text`,
 t.`retweet_count` as `retweet_count`,t.`in_reply_to_user` as `in_reply_to_user`,t.`id` as `id`,t.`coordinate` as `coordinate`,t.`in_reply_to_status` as `in_reply_to_status`,t.`user`.`status_count` as `user.status_count`,
 t.`geo_tag`.`stateID` as `geo_tag.stateID`,t.`create_at` as `create_at`,t.`lang` as `lang`,t.`hashtags` as `hashtags`
 from twitter.ds_tweet t) ta
where ta.lang_len ...
```
However, the current SQLPPGenerator didn't calculate the projected fields correctly, which cased the query returns empty result.

To handle this, I used a much cleaner way to avoid project all fields of the source table, but instead:
```
select ...
from (select length(lang) as `lang_len`,t
          from twitter.ds_tweet t) ta
where ta.`lang_len` ... AND ta.t.`create_at`...
```

### DIFF
--- a/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
+++ b/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
@@ -144,16 +144,17 @@ class SQLPPGenerator extends AsterixQueryGenerator {
       ParsedResult(Seq.empty, exprMap)
     } else {
       val producedExprs = mutable.LinkedHashMap.newBuilder[String, FieldExpr]
+      
       appends.foreach { append =>
         val as = append.as
-        producedExprs += append.as.name -> FieldExpr(s"$appendVar.${as.name}", append.definition)
+        producedExprs += append.as.name -> FieldExpr(s"$appendVar.$quote${as.name}$quote", append.definition)
       }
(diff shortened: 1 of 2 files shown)


HotLemonJuice:
LGTM

PR #341, opened 2017-05-06
Update the script to not drop the UDF

JavierJia:
The UDF is also under the `twitter` dataverse. So it will drop the pre-existing UDF. 
This PR remove this line.


HotLemonJuice:
LGTM

JavierJia:
@HotLemonJuice since your new PR include this line, I'm closing this.

Issue #343, opened 2017-05-08
Zoom-in and move the map region before a search prevents showing the results.


waans11:
Open the site and zoom-in and move the region. Then do a search. It will not show any results.

PR #349, opened 2017-05-12
Fix Zoom-in No Query Results

HotLemonJuice:
Fix issue #343 

Main idea:
- Basically, we need to reset the geoIDs whenever the map is zoomed, not just after we did a query.
- Because we just Ajax call to get polygons, which is a necessary information to reset geoIDs, I added one more statement inside the Ajax call to reset the geoIDs whenever it is done.

Miscellaneous: 
- some refactoring in the service.js. Asterix => cloudberry


twittermap/public/javascripts/map/controllers.js
```diff
@@ -90,6 +90,13 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
       });
     }
 
+    function resetGeoInfo(level) {
+      $scope.status.logicLevel = level;
+      if ($scope.geojsonData[level])
+        resetGeoIds($scope.bounds, $scope.geojsonData[level], level + 'ID');
+      cloudberry.parameters.geoLevel = level;
```
JavierJia: a picky comment, 

it feels very unnature to have two lines before and after the `if` clause. I was thinking `cloudberry.parameters.geoLevel = level;` is inside the if. 

It will be better to group the non-branch statements together. 
```
   $scope.status.logicLevel = level;
   cloudberry.parameters.geoLevel = level;
   if ($scope.geojsonData[level])
       resetGeoIds($scope.bounds, $scope.geojsonData[level], level + 'ID');
```

or put it all after the if. you decide :-)

PR #351, opened 2017-05-13
Add commands to restart docker after installation

luochen01:


docs/quick-start.md
```diff
@@ -36,6 +36,12 @@ The second command will download and run a prebuilt AsterixDB docker container f
 After it finishes, you should see the messages as shown in the following screenshot:
 ![docker][docker]
 
+After you have installed the docker container for the first time, you may restart it later using the following commands:
+
+```
+~> docker start cc
+~> docker start nc1
```
JavierJia: we can make it one line:
```
docker start cc nc1 
```

And probably also add the stop script? 
```
docker stop cc nc1
```

luochen01: Done.

HotLemonJuice:
👍

docs/quick-start.md
```diff
@@ -36,6 +36,18 @@ The second command will download and run a prebuilt AsterixDB docker container f
 After it finishes, you should see the messages as shown in the following screenshot:
 ![docker][docker]
 
+After you have installed the docker container for the first time, you may restart it later using the following command:
```
JavierJia: I still have a feeling that this part is just a "good to know" information, which doesn't serve the purpose of "quick-start". we can also tell them how to start and stop the AsterixDB, but I don't know if it's this page's task. 

I think the developer need to know this command, it should go to wiki?

luochen01: For an end-user of Cloudberry, it is still common that he/she installed the AsterixDB/TwitterMap for the first time. After the machine is shutdown, what should be done then? Definitely he shouldn't re-install docker/AsterixDB again. I think such information is needed, but maybe put it at other places should also be fine.

chenlica: It's good to have these instructions, but we need to make it very clear when they are needed.  How about the following:

"In case the docker container is shut down accidentally, you can restart it using the following command:"

luochen01: Got it. Still a few questions.
1. Do we need to include instructions on how to shutdown docker?
2. I think 'accidentally' it not a very proper word. For example, the user may restart his machine, and after that he wants to run the TwitterMap again. Restarting the machine perhaps shouldn't be called as 'accidentally shutdown'.
What about:
"After you have installed the docker container for the first time, you can restart it using the following command in case the docker is shutdown:"

chenlica: 1. Yes, please add it.
2. Please rephrase it as you suggested.

luochen01: Done.

chenlica:
LGTM now.

Issue #338, opened 2017-05-06
Cloudberry should supports static dataset


haochen07:
Cloudberry now only supports dataset that has a `timeField`. For some of the datasets where information is rather static and mainly used for lookups, the way we handle them now is to append a fake `timeField` in the schema, which is not elegant and fails to return `Stats` information.
We need to distinguish `dynamic dataset` and `static dataset`, and different `Stats` respectively, and treat them in proper way.

chenlica:
Per our discussion, please add our proposed plan to this issue for our records.

haochen07:
- Add a field `type` to ddl, refer to `dynamic` dataset or `static` dataset.
- Dynamic datasets are required to assign a `timeField`, while in Static datasets it is optional. `timeField` in Static datasets will be treated no different as other dimensions.
- For static datasets:
  - In `QueryPlanner`, all queries will be executed directly to database, no views.
  - Static datasets will ignore `suggestViews` logic.
  - In `QuerySlicing`, do not slice.
- Design new trait `IStats`. Previous `Stats` will be `dynamicStats`. `StaticStats` will get rid of all `timeField` things and support `cardinality` for the moment.
- In `Register`, collect `IStats` and append it to `meta` table.

haochen07:
@chenlica @JavierJia You can have a review on it.

chenlica:
The plan LGTM.

JavierJia:
👍

PR #356, opened 2017-05-21
[Issue #338] Lookup dataset

haochen07:
This PR related to #338.

**Problem**
- `timeField` is currently required for all schemas, whereas for lookup tables they may not have such a time field. We need to create a fake `timeField` in their schema to put them in cloudberry.
- When collecting `Stats` information in register process, since `timeField` is fake for lookup tables, their collection process will always fail, which make it impossible to distinguish real failure cases for dynamic tables.

**Implementation** 
- Make `TemporalSchema` and `LookupSchema` inherit `Schema`. `timeField` only exists in `TemporalSchema`.
- In `UnresolvedSchema`, `timeField` is marked as optional. So in schema json, this field can be absent or maked as `null`. When resolving a schema, the one that has `timeField` will be converted to `TemporalSchema` and those without `timeField` will be converted as `LookupSchema`.
- Change tweets schema to `temporal` and other population schema to `lookup`.
- `LookupSchema` now only support lookup queries.
- When collecting `Stats` in register, `TemporalSchema` will do the `Stats` queries and catch possible exceptions whereas `LookupSchema` will append fake `Stats` because for the moment, the `Stats` for `LookupSchema` is not useful.
- Add test cases specifically for `Lookup` dataset and change register json of lookup tables.


haochen07:
The codebase is finished and previous test cases are updated in order to fit new changes. Need more test cases on `StaticSchema`. You can now have a review on the codebase changes first. @JavierJia @luochen01

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -54,10 +54,16 @@ class BerryClient(val jsonParser: JSONParser,
     case initial: Initial if initial.ts == curKey =>
       val queryInfos = initial.queries.map { query =>
         val info = initial.infos(query.dataset)
-        val bound = query.getTimeInterval(info.schema.timeField).getOrElse(new TInterval(info.dataInterval.getStart, DateTime.now))
+        val timeField = info.schema match {
```
luochen01: I think it's better to make this code block as a method for schema?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -274,17 +295,24 @@ class DataStoreManager(metaDataset: String,
...
+        val cardinalityQuery = Query(dataset, globalAggr = Some(GlobalAggregateStatement(AggregateStatement(schema.fieldMap("*"), Count, Field.as(Min(timeField), "count")))))
+        val parser = queryGenFactory()
+        import TimeField.TimeFormat
+        for {
+          minTime <- conn.postQuery(parser.generate(minTimeQuery, Map(dataset -> schema))).map(r => (r \\ "min").head.as[String])
+          maxTime <- conn.postQuery(parser.generate(maxTimeQuery, Map(dataset -> schema))).map(r => (r \\ "max").head.as[String])
+          cardinality <- conn.postQuery(parser.generate(cardinalityQuery, Map(dataset -> schema))).map(r => (r \\ "count").head.as[Long])
+        } yield (new TJodaInterval(TimeFormat.parseDateTime(minTime), TimeFormat.parseDateTime(maxTime)), cardinality)
+      case static: StaticSchema =>
+        throw new Exception("Cannot collect Stats information for static dataset " + static.typeName)
```
luochen01: Make this Exception more specific.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -63,14 +63,21 @@ class OriginalDataAgent(val dataSetInfo: DataSetInfo,
...
   private def collectStats(start: DateTime): Unit = {
-    val now = DateTime.now().minusMillis(1)
-    val filter = FilterStatement(schema.timeField, None, Relation.inRange, Seq(start, now).map(TimeField.TimeFormat.print))
-    val aggr = GlobalAggregateStatement(AggregateStatement(schema.fieldMap("*"), Count, Field.as(Count(schema.fieldMap("*")), "count")))
-    val queryCardinality = Query(dbName, filter = Seq(filter), globalAggr = Some(aggr))
-
-    conn.postQuery(queryParser.generate(queryCardinality, Map(dbName -> schema)))
-      .map(r => new Cardinality(start, now, (r \\ "count").head.as[Long]))
-      .pipeTo(self)
+    schema match {
```
luochen01: where does this schema come from?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -31,14 +31,21 @@ class AQLGenerator extends IQLGenerator {
   //TODO combine with parseQuery
   override def calcResultSchema(query: Query, schema: Schema): Schema = {
     if (query.lookup.isEmpty && query.groups.isEmpty && query.select.isEmpty) {
-      schema.copy()
+      schema match {
```
luochen01: I think you can make copy as a method of schema.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -31,14 +31,21 @@ class AQLGenerator extends IQLGenerator {
...
     } else {
       ???
     }
   }
 
   def parseCreate(create: CreateView, sourceSchema: Schema): String = {
-    val resultSchema = calcResultSchema(create.query, sourceSchema)
+    val resultSchema = calcResultSchema(create.query, sourceSchema) match {
+      case temporal: TemporalSchema => temporal
+      case static: StaticSchema => throw new Exception("Create View cannot be applied for static dataset " + static.typeName)
```
luochen01: I think this part should be checked by here. The caller needs to ensure this.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -351,8 +358,12 @@ class AQLGenerator extends IQLGenerator {
     val fields = schema.fieldMap.values.filter(f => f.dataType != DataType.Hierarchy && f != AllField).map {
       f => mkNestDDL(f.name.split("\\.").toList, fieldType2ADMType(f) + (if (f.isOptional) "?" else ""))
     }
+    val schemaTypeName = schema match {
+      case temporal: TemporalSchema => temporal.typeName
```
luochen01: No need to distinguish them.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
```diff
@@ -130,7 +130,10 @@ abstract class AsterixQueryGenerator extends IQLGenerator {
 
   def calcResultSchema(query: Query, schema: Schema): Schema = {
     if (query.lookup.isEmpty && query.groups.isEmpty && query.select.isEmpty) {
-      schema.copy()
+      schema match {
+        case temporal: TemporalSchema => temporal.copy()
```
luochen01: Same here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
```diff
@@ -283,8 +286,12 @@ abstract class AsterixQueryGenerator extends IQLGenerator {
     val fields = schema.fieldMap.values.filter(f => f.dataType != DataType.Hierarchy && f != AllField).map {
       f => mkNestDDL(f.name.split("\\.").toList, fieldType2ADMType(f) + (if (f.isOptional) "?" else ""))
     }
+    val schemaTypeName: String = schema match {
+      case temporal: TemporalSchema => temporal.typeName
```
luochen01: Same here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/DataSetInfo.scala
```diff
@@ -45,14 +46,18 @@ object DataSetInfo {
 
         val schema = dataSetInfo.schema
         val primaryKey = schema.primaryKey.map(schema.getField(_).get)
-        val timeField = schema.getField(schema.timeField) match {
-          case Some(f) if f.isInstanceOf[TimeField] => f.asInstanceOf[TimeField]
-          case None if schema.timeField.isEmpty => throw new QueryParsingException(s"Time field is not specified for ${schema.typeName}.")
-          case _ => throw new QueryParsingException(s"${schema.timeField} is not a valid time field.")
+        val resolvedSchema = schema.timeField match {
+          case Some(field) =>
```
luochen01: I think I saw the exactly same code snippet before. Could we unify them as a single method?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
```diff
@@ -30,20 +30,26 @@ object QueryResolver {
...
-    val (filter, fieldMapAfterFilter) = resolveFilters(query.filter, fieldMapAfterUnnest)
-
-    val (groups, fieldMapAfterGroup) = resolveGroup(query.groups, fieldMapAfterFilter, schemaMap)
-    val (select, fieldMapAfterSelect) = resolveSelect(query.select, fieldMapAfterGroup)
-
-    val (globalAggr, fieldMapAfterGlobalAggr) = resolveGlobalAggregate(query.globalAggr, fieldMapAfterSelect)
-
-    Query(query.dataset, append, lookup, filter, unnest, groups, select, globalAggr, query.estimable)
+    schemaMap(query.dataset) match {
+      case temporal: TemporalSchema =>
```
luochen01: Same here. It's better to reduce the code of checking temporal/static schemas.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
```diff
@@ -30,20 +30,26 @@ object QueryResolver {
...
+        val (unnest, fieldMapAfterUnnest) = resolveUnnests(query.unnest, fieldMapAfterLookup)
+        val (filter, fieldMapAfterFilter) = resolveFilters(query.filter, fieldMapAfterUnnest)
+        val (groups, fieldMapAfterGroup) = resolveGroup(query.groups, fieldMapAfterFilter, schemaMap)
+        val (select, fieldMapAfterSelect) = resolveSelect(query.select, fieldMapAfterGroup)
+        val (globalAggr, fieldMapAfterGlobalAggr) = resolveGlobalAggregate(query.globalAggr, fieldMapAfterSelect)
+
+        Query(query.dataset, append, lookup, filter, unnest, groups, select, globalAggr, query.estimable)
+
+      case static: StaticSchema =>
+        val fieldMap = static.fieldMap
```
luochen01: This is not correct. Why would a query refers a static schema, but only performs append/lookup operations?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLPPGenerator.scala
```diff
@@ -70,7 +70,10 @@ class SQLPPGenerator extends AsterixQueryGenerator {
 
   def parseCreate(create: CreateView, schemaMap: Map[String, Schema]): String = {
     val sourceSchema = schemaMap(create.query.dataset)
-    val resultSchema = calcResultSchema(create.query, schemaMap(create.query.dataset))
+    val resultSchema = calcResultSchema(create.query, sourceSchema) match {
```
luochen01: same here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/unresolved.scala
```diff
@@ -21,13 +21,24 @@ object Unresolved {
...
   def toUnresolved(schema: Schema): UnresolvedSchema = {
-    UnresolvedSchema(
-      schema.typeName,
-      schema.dimension,
-      schema.measurement,
-      schema.primaryKey.map(_.name),
-      schema.timeField.name
-    )
+    schema match {
+      case temporal: TemporalSchema =>
```
luochen01: Same here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Schema.scala
```diff
@@ -184,19 +184,31 @@ trait Measurement {
...
-                  dimension: Seq[Field],
-                  measurement: Seq[Field],
-                  primaryKey: Seq[Field],
-                  timeField: TimeField
-                 ) {
+abstract class Schema(typeName: String,
+                      dimension: Seq[Field],
+                      measurement: Seq[Field],
+                      primaryKey: Seq[Field]
+                     ) {
```
luochen01: I think it's better to think a little bit how to unify the common parts of TemporalSchema and StaticSchema. Try to reduce the distinguish of them in your codebase, but rather unify them as the abstract schema.
For example, is it OK to add `timeField` as a method of Schema, but only TemporalSchema implements it? Calling `timeField` on StaticSchema would automatically throw an Exception.

zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TwitterDataStore.scala
```diff
@@ -5,34 +5,35 @@ import edu.uci.ics.cloudberry.zion.model.schema._
...
-                                           TextField("text"),
-                                           NumberField("in_reply_to_status"),
-                                           NumberField("in_reply_to_user"),
-                                           NumberField("favorite_count"),
-                                           NumberField("retweet_count"),
-                                           NumberField("user.status_count")
-                                         ),
-                                         Seq(NumberField("id")),
-                                         TimeField(TimeFieldName))
+  val TwitterSchema = TemporalSchema(
```
luochen01: Code Format Issue here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
haochen07: What we want is to only enable `staticSchema` do `lookup` queries. Then what may be the right way to go?

haochen07:
Test cases for static dataset added and register json format for population tables are changed.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -133,47 +138,38 @@ class DataStoreManager(metaDataset: String,
...
-            sender ! DataManagerResponse(true, "Register Finished: dataset " + dataSetName + " has successfully registered.\n")
 
-          case Failure(_) =>
-            //FIXME Need to introduce static dataset to handle this
+          case static: StaticSchema =>
             val currentDateTime = new DateTime()
-            val fakeStats = Stats(currentDateTime, currentDateTime, currentDateTime, 1000000)
-            val fakeStartDate = new DateTime(2005, 3, 26, 12, 0, 0, 0)
+            val fakeStats = Stats(currentDateTime, currentDateTime, currentDateTime, 1000)
+            val fakeStartDate = new DateTime(2017, 1, 1, 12, 0, 0, 0)
```
luochen01: What is the purpose of this fakeTimestamp? Could we ignore the stat info for Static datasets?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
luochen01: What you want to do is that you can lookup a static dataset, instead of perform a lookup query for a static dataset.
I suggest it's better to keep the query resolver as it is, and move such logic to QueryValidator (which is used to validate a query).

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
haochen07: If no `Stats` info, we need to change `DataSetInfo`, which will require more relative changes and reconsider logics behind. 
The purpose of fake `Stats` is 1) to maintain the structure of `DataSetInfo`; 2) `cardinality` in `Stats` is number of records in the table, which may be useful in the future (although it is fake now)

haochen07:
The logic is transferred from `QueryResolver` to `QueryValidator`.

haochen07:
Register jsons are changed so that now `population` datasets are `static`. 
This PR is all done and ready for final check.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Schema.scala
JavierJia: IMO,  if we need to throw some exception in the implementation, that means the interface design is not correct. The `timeField` by nature is the property of the `Temporal` schema. 

In most of the code after query parsing, the parameter should be of type `TemporalSchema` only if it relies on the `timeField`.

haochen07: Yes. The thing is that, if to make a flawless design after introducing static schema, we need to change the `DataSetInfo` and then make previous function calls only available for temporal schema or `DatasetInfo` who contains a temporal dataset. Need to consider carefully.

haochen07:
@luochen01 Maybe a final check for this PR?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryValidator.scala
```diff
@@ -46,7 +46,11 @@ object QueryValidator {
     t.isInstanceOf[Number] || implicitly[TypeTag[T]].tpe <:< typeOf[AnyVal]
   }
 
-  def validateQuery(query: Query): Unit = {
+  def validateQuery(query: Query, schemaMap: Map[String, Schema]): Unit = {
+    if (!schemaMap(query.dataset).isTemporal) {
+      Logger.error("static dataset " + query.dataset + " does not support query " + query.toString)
```
luochen01: Throw an exception here.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -78,25 +78,25 @@ class DataStoreManager(metaDataset: String,
...
-                  } else {
-                    val now = DateTime.now()
-                    val compensate = FilterStatement(temporal.timeField, None, Relation.inRange,
-                      Seq(info.stats.lastModifyTime, now).map(TimeField.TimeFormat.print))
-                    val appendQ = createQuery.copy(filter = compensate +: createQuery.filter)
-                    answerQuery(AppendView(info.name, appendQ), Some(now))
-                  }
-                case None => log.warning(s"can not append to a base dataset: $append.dataset.")
+          if (!info.schema.isTemporal) {
+            log.error("Append View operation cannot be applied to static dataset " + info.name)
```
luochen01: Should add a return here?

haochen07: No. we can't do that in a `Receive` function. Change the logic to `if-else`.

JavierJia:
Since we treat the `lookup` dataset separately, we may not need to emphasize that a dataset is `Temporal` which is the default setting.

haochen07:
Yes, then what maybe the change? I think the logic is the same.

JavierJia:
maybe just remove `Temporal` prefix?

haochen07:
I think it is better to have a name other than `Schema`, because `Schema` is reserved to the abstract class name.

JavierJia:
Then maybe we can rename it as `AbstractSchema`?

haochen07:
OK. That would work.

haochen07:
Done.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -141,7 +141,7 @@ class DataStoreManager(metaDataset: String,
     if(!metaData.contains(dataSetName)){
       try{
         dataSetRawSchema.toResolved match {
-          case temporal: TemporalSchema =>
+          case temporal: Schema =>
```
JavierJia: also update the variable name?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -63,11 +63,11 @@ class OriginalDataAgent(val dataSetInfo: DataSetInfo,
   }
 
   private def collectStats(start: DateTime): Unit = {
-    if (!dataSetInfo.schema.isTemporal) {
+    if (!dataSetInfo.schema.hasTimeField) {
       log.error("Cannot do aggregation query for static dataset " + dataSetInfo.schema.getTypeName)
```
JavierJia: `for static dataset` --> `for lookup datasets`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryValidator.scala
```diff
@@ -46,8 +46,8 @@ object QueryValidator {
     t.isInstanceOf[Number] || implicitly[TypeTag[T]].tpe <:< typeOf[AnyVal]
   }
 
-  def validateQuery(query: Query, schemaMap: Map[String, Schema]): Unit = {
-    if (!schemaMap(query.dataset).isTemporal) {
+  def validateQuery(query: Query, schemaMap: Map[String, AbstractSchema]): Unit = {
+    if (!schemaMap(query.dataset).hasTimeField) {
       throw new IllegalArgumentException("static dataset " + query.dataset + " does not support query " + query.toString)
```
JavierJia: `static dataset` --> `lookup dataset`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -199,26 +199,26 @@ abstract class Schema(typeName: String,
 
   def getTypeName: String = typeName
   def getTimeField: TimeField
-  def copySchema: Schema
-  def asTemporal: TemporalSchema
-  def isTemporal: Boolean
+  def copySchema: AbstractSchema
+  def asSchema: Schema
```
JavierJia: I still feel `asSchema` should not be a method of the base class. if needed, we can always convert it using `asInstanceOf`

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -199,26 +199,26 @@ abstract class Schema(typeName: String,
 
   def getTypeName: String = typeName
   def getTimeField: TimeField
```
JavierJia: what could be the result if the `hasTimeField` is false?

haochen07: Since now the core difference between `Schema` and `LookupSchema` is the existence of `timeField`, then if it is false, we could know that this is a `LookupSchema`.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
haochen07: Yes, we can use `hasTimeField` and `asInstanceOf` to do the same thing as `asSchema`. Will remove it, but the logic is the same.

haochen07:
Naming issue fixed and `asSchema` is removed. Wait for further comments.

haochen07:
Personally, I think to resolve the issue you concerned about, it would be good to differentiate different `DataSetInfo` with respect to different type of `AbstractSchema`, i.e. `LookupDataSetInfo` etc. However, this will influence even more part of the codebase than the current PR, and should take a careful consideration on it.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
JavierJia: if that is the case, I think we should remove the `hasTimeField` method, and let `getTimeField` returns an `Option[Timefield]`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -38,7 +38,10 @@ class AQLGenerator extends IQLGenerator {
   }
 
   def parseCreate(create: CreateView, sourceSchema: AbstractSchema): String = {
-    val resultSchema = calcResultSchema(create.query, sourceSchema).asSchema
+    if (!sourceSchema.hasTimeField){
```
JavierJia: since it won't work for `lookupSchema`, can we just change the signature from `sourceSchema: AbstractSchema` to `sourceSchema: Schema` ?

haochen07:
I didn't change for `parseUpsert`, `parseDelete`, `parseDrop` as I am not sure if they are useful for  upsert, delete and drop lookup dataset.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/AbstractDataSetAgent.scala
```diff
@@ -21,7 +21,7 @@ import scala.concurrent.{ExecutionContext, Future}
   * @param ec
   */
 abstract class AbstractDataSetAgent(val dbName: String,
-                                    val schema: Schema,
+                                    val schema: AbstractSchema,
```
JavierJia: shoud this be of type `Schema`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/AbstractDataSetAgent.scala
```diff
@@ -62,7 +62,7 @@ abstract class AbstractDataSetAgent(val dbName: String,
 }
 
 abstract class AbstractUpdatableDataSetAgent(override val dbName: String,
-                                             override val schema: Schema,
+                                             override val schema: AbstractSchema,
```
JavierJia: ditto?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -78,20 +78,26 @@ class DataStoreManager(metaDataset: String,
...
-                val appendQ = createQuery.copy(filter = compensate +: createQuery.filter)
-                answerQuery(AppendView(info.name, appendQ), Some(now))
-              }
-            case None => log.warning(s"can not append to a base dataset: $append.dataset.")
+          if (info.schema.getTimeField.isEmpty) {
+            log.error("Append View operation cannot be applied to lookup dataset " + info.name)
+          } else {
+            info.createQueryOpt match {
+              case Some(createQuery) =>
+                if (createQuery.filter.exists(_.field == info.schema.getTimeField)) {
```
JavierJia: `info.schema.getTimeField` returns an option now, right?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -245,11 +243,15 @@ class DataStoreManager(metaDataset: String,
       log.warning(s"invalid dataset in the CreateView msg: $create")
       return
     }
-    creatingSet.add(create.dataset)
```
JavierJia: why remove this line?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -245,11 +243,15 @@ class DataStoreManager(metaDataset: String,
...
     val sourceInfo = metaData(create.query.dataset)
-    val schema = managerParser.calcResultSchema(create.query, sourceInfo.schema)
+    if (sourceInfo.schema.getTimeField.isEmpty) {
+      log.error("Create View cannot be applied for lookup dataset " + sourceInfo.schema.getTypeName)
+      return
+    }
+    creatingSet.add(create.dataset)
+    val schema = managerParser.calcResultSchema(create.query, sourceInfo.schema.asInstanceOf[Schema])
```
JavierJia: it's already a `Schema` right?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -301,7 +307,7 @@ object DataStoreManager {
     val View = Value("view")
   }
 
-  type ChildMakerFuncType = (AgentType.Value, ActorRefFactory, String, String, Schema, Option[DataSetInfo], IQLGenerator, IDataConn, Config) => ActorRef
+  type ChildMakerFuncType = (AgentType.Value, ActorRefFactory, String, String, AbstractSchema, Option[DataSetInfo], IQLGenerator, IDataConn, Config) => ActorRef
```
JavierJia: should it also be `Schema`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -315,7 +321,7 @@ object DataStoreManager {
                    context: ActorRefFactory,
                    actorName: String,
                    dbName: String,
-                   dbSchema: Schema,
+                   dbSchema: AbstractSchema,
```
JavierJia: ditto?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ViewDataAgent.scala
```diff
@@ -3,13 +3,13 @@ package edu.uci.ics.cloudberry.zion.actor
...
 import edu.uci.ics.cloudberry.zion.model.datastore.{IDataConn, IQLGenerator}
-import edu.uci.ics.cloudberry.zion.model.schema.{AppendView, Query, Schema}
+import edu.uci.ics.cloudberry.zion.model.schema.{AppendView, Query, AbstractSchema}
 import play.api.libs.json.JsValue
 
 import scala.concurrent.ExecutionContext
 
 class ViewDataAgent(override val dbName: String,
-                    override val schema: Schema,
+                    override val schema: AbstractSchema,
```
JavierJia: this should also be `Schema`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ViewDataAgent.scala
```diff
@@ -27,7 +27,7 @@ class ViewDataAgent(override val dbName: String,
 }
 
 object ViewDataAgent {
-  def props(dbName: String, schema: Schema, queryParser: IQLGenerator, conn: IDataConn, config: Config)
+  def props(dbName: String, schema: AbstractSchema, queryParser: IQLGenerator, conn: IDataConn, config: Config)
```
JavierJia: ditto.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -31,12 +32,31 @@ class AQLGenerator extends IQLGenerator {
...
   override def calcResultSchema(query: Query, schema: Schema): Schema = {
     if (query.lookup.isEmpty && query.groups.isEmpty && query.select.isEmpty) {
-      schema.copy()
+      schema.copySchema
     } else {
       ???
     }
   }
 
+  private def splitSchemaMap(schemaMap: Map[String, AbstractSchema]): (Map[String, Schema], Map[String, LookupSchema]) = {
```
JavierJia: we saw this function three times :-).  it should indeed be a static function.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
```diff
@@ -122,21 +123,40 @@ abstract class AsterixQueryGenerator extends IQLGenerator {
...
     if (query.lookup.isEmpty && query.groups.isEmpty && query.select.isEmpty) {
-      schema.copy()
+      schema.copySchema
     } else {
       ???
     }
   }
 
-  protected def initExprMap(dataset: String, schemaMap: Map[String, Schema]): Map[String, FieldExpr] = {
+  private def splitSchemaMap(schemaMap: Map[String, AbstractSchema]): (Map[String, Schema], Map[String, LookupSchema]) = {
```
JavierJia: 4th ...

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -184,20 +185,51 @@ trait Measurement {
...
                   dimension: Seq[Field],
                   measurement: Seq[Field],
                   primaryKey: Seq[Field],
                   timeField: TimeField
-                 ) {
+                 ) extends AbstractSchema(typeName, dimension, measurement, primaryKey){
 
-  private val dimensionMap: Map[String, Field] = dimension.map(f => f.name -> f).toMap
-  private val measurementMap: Map[String, Field] = measurement.map(f => f.name -> f).toMap
+  override def getTimeField: Option[TimeField] = Some(timeField)
```
JavierJia: so here we have the `timeField` and also the `getTimeField` function. 
Can we further remove the `getTimeField` by checking the instance of the `AbstractSchema` by using `isInstanceOf[Schema]`? 

It has two benefits comparing to use `getTimeField` function. 
1. it's more general, usually, it feels unnature to decide the type of the object by checking the function return values.
2. after conversion, we can use `timeField` instead of `getTimeField.get` which is error prone.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
haochen07: I followed the logic of `calcResultSchema` to duplicate the code twice in both `AQLGenerator` and `AsterixQueryGenerator`. I didn't find a proper object to apply this static method. Which one you think maybe appropriate?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
```diff
@@ -91,7 +108,7 @@ class AQLGenerator extends IQLGenerator {
       }
     }
     
-    val varMapAfterLookup = parseLookup(query.lookup, schemaVars, schemaMap)
+    val varMapAfterLookup = parseLookup(query.lookup, schemaVars, lookupSchemaMap)
```
haochen07: Actually not. I saw line 170 of `AQLGenerator` and this is the only place to use this schemaMap and actually this line is useless. I just doubt so I didn't change it. Can I remove line 170 in `AQLGenerator`?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
haochen07: Strange. I only saw twice..

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
JavierJia: If it's useless, let's remove it.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AQLGenerator.scala
JavierJia: I think `AsterixQueryGenerator` should be good.
```
object AsterixQueryGenerator {
   def splitSchemaMap()...
}
```

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/AbstractDataSetAgent.scala
haochen07: `LookupSchema` doesn't have any `dataAgent` at all right?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
haochen07: Not removed. See 251

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ViewDataAgent.scala
haochen07: For all the logic about `dataSetAgent`. I need to check for the logic and give you possibility to do that.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/AbstractDataSetAgent.scala
JavierJia: no, it doesn't.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
haochen07: I suppose `isInstanceOf[Schema]` is actually the same as `hasTimeField`, which we removed previously?

JavierJia: yes. here I mean the `getTimeField` can also be removed as well. 

if we want to detect if the schema is temporal, using `isInstanceof`. 
if we want to access the timefield, convert it to `val schema = abs.asInstanceof[Schema]` and call `schema.timeField`.  IMO it's more clear.

haochen07:
Updated. Not sure about how to deal with some of the exceptions.

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -54,9 +54,13 @@ class BerryClient(val jsonParser: JSONParser,
     case initial: Initial if initial.ts == curKey =>
       val queryInfos = initial.queries.map { query =>
         val info = initial.infos(query.dataset)
-        val bound = query.getTimeInterval(info.schema.getTimeField.get).getOrElse(new TInterval(info.dataInterval.getStart, DateTime.now))
+        if (!info.schema.isInstanceOf[Schema]) {
+          ??? // How to deal with this?
```
JavierJia: throw an exception?

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -150,7 +154,11 @@ class BerryClient(val jsonParser: JSONParser,
     val futures = Future.traverse(queryGroup.queries) { queryInfo =>
       if (queryInfo.queryBound.overlaps(interval)) {
         val overlaps = queryInfo.queryBound.overlap(interval)
-        val timeFilter = FilterStatement(queryInfo.dataSetInfo.schema.getTimeField.get, None, Relation.inRange,
+        if (!queryInfo.dataSetInfo.schema.isInstanceOf[Schema]){
+          ???
```
JavierJia: if we have previously checked it already, we should not do it again? (especially for the private function.)

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -78,16 +78,17 @@ class DataStoreManager(metaDataset: String,
...
         case Some(info) =>
-          if (info.schema.getTimeField.isEmpty) {
+          if (!info.schema.isInstanceOf[Schema]) {
             log.error("Append View operation cannot be applied to lookup dataset " + info.name)
           } else {
+            val schema = info.schema.asInstanceOf[Schema]
             info.createQueryOpt match {
               case Some(createQuery) =>
-                if (createQuery.filter.exists(_.field == info.schema.getTimeField)) {
+                if (createQuery.filter.exists(_.field == schema.timeField)) {
```
JavierJia: much better :-)

zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManager.scala
```diff
@@ -217,7 +218,10 @@ class DataStoreManager(metaDataset: String,
 
     val actor = context.child("data-" + query.dataset).getOrElse {
       val info = metaData(query.dataset)
-      val schema: AbstractSchema = info.schema
+      if (!info.schema.isInstanceOf[Schema]) {
+        ???
```
JavierJia: throw an exception?

PR #358, opened 2017-05-22
Add the "drugmap" demo to TwitterMap

JavierJia:
This PR add the "drugmap" page to the TwitterMap UI under `/drugmap` path. 

1. This page will use the fixed `dropbox` option to only let users choose "drug" as the keyword.
2. Add some configurations to make the label text configurable.


chenlica:
@JavierJia Please add a screenshot to the PR.  Let's do a F2F meeting to do the review.

twittermap/public/javascripts/map/controllers.js
```diff
@@ -671,7 +671,7 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common'])
           if($scope.doSentiment) {
             $scope.infoPromp = "Score";  // change the info promp
```
HotLemonJuice: Since you are using config to set the prompt, I think maybe you can remove this line.

twittermap/app/views/twittermap/main.scala.html
```diff
@@ -24,6 +24,14 @@
       wsURL:"@ws",
       sentimentEnabled: @sentimentEnabled,
       sentimentUDF: "@sentimentUDF",
+      startDate: new Date("@startDate"),
+      @if(isDrugMap) {
+        mapLegend: "$",
+        sumText : "dollars",
+      } else {
+        mapLegend: "Count:",
```
HotLemonJuice: Can you add `Score` here if `sentimentEnabled` is true?

JavierJia: I think the sentiment related logic has covered this case. I don't want to make this half scala half html file too complicated.

twittermap/public/javascripts/map/controllers.js
JavierJia: hmm, I think we still need to. `sentiment` is not always enabled. If user toggle off the `sentiment`, we will still need to reset it. One `config.mapLegend` is not enough.

JavierJia:
Here is the screenshot of the `/drugmap` page:
screen shot 2017-05-24 at 10 48 58 am
screen shot 2017-05-24 at 10 49 04 am

JavierJia:
@HotLemonJuice per our discussion with @chenlica, we are going to merge this PR. Any objections before I merge?

HotLemonJuice:
👍

chenlica:
@JavierJia I know this PR has been merged.  Can you change "keywords list" to "keywords"?

JavierJia:
OK.

PR #367, opened 2017-06-16
change authors of the new paper

HotLemonJuice:
Change: List only Shengjie Xu as the author of the new paper


JavierJia:
did you also update the content in the paper?

HotLemonJuice:
@JavierJia No, in my paper I only listed myself as the author, so there is no need to update it.

JavierJia:
LGTM 👍

PR #376, opened 2017-07-13
IE alert

hh1680651:
Add an alert bar when using IE browser:
alert-bar


JavierJia:
Since there are already many comments in the old PR, let's do the feature branch thing in further? Can you close this PR and keep working on the old one?

Issue #372, opened 2017-07-05
Support SQL backend


chenlica:
Currently Cloudberry supports AsterixDB as its backend using a translator to SQL++.  To make Cloudberry more general, we want to support other backends as well.  

In https://github.com/ISG-ICS/cloudberry/issues/287 we finished the step of supporting SparkSQL, and its core is really SQL.  In this issue, we want to fully support SQL on top of the previous PR.  Specifically:

Step 1: Support write operators using SQL;
Step 2: Setup a TwitterMap using MySQL as the backend.

hh1680651:
![outline](https://user-images.githubusercontent.com/23527044/27887728-161f4418-6197-11e7-97df-931d4c332194.jpg)

JavierJia:
Here are steps based on our discussion
* [ ] insert the data into the MySQL db. 
* [ ] summarize the differences between the SparkSQL with the MySQL queries. Depend on the number of differences, we will decide the design strategy.

PR #378, opened 2017-07-20
MySQL parseQuery

hh1680651:
Almost fully support MySQL backend.


cloudberry/neo/app/controllers/Cloudberry.scala
```diff
@@ -39,6 +39,7 @@ class Cloudberry @Inject()(val wsClient: WSClient,
   val (asterixConn, qlGenerator) =
     config.AsterixLang match {
       case "sparksql" => (new SparkConn(url = "local[4]"), SparkSQLGenerator)
+      case "sql" => (new SQLConn(url = "jdbc:mysql://localhost:3306/twitter_ds_tweet"), SQLGenerator)
```
JavierJia: make it configurable like the `sqlpp` one?

cloudberry/project/dependencies.scala
```diff
@@ -26,6 +26,9 @@ object Dependencies {
   val zionDependencies: Seq[ModuleID] = Seq(
     "joda-time" % "joda-time" % "2.9.3",
     "com.typesafe.akka" %% "akka-actor" % "2.4.4",
+    "org.scalikejdbc" %% "scalikejdbc" % "2.5.2",
+    "mysql" % "mysql-connector-java" % "5.1.24",
+//    "com.typesafe.play" %% "anorm" % "2.5.1",
```
JavierJia: remove unused code?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -13,6 +13,8 @@ class JSONParser extends IJSONParser {
   import JSONParser._
 
   override def getDatasets(json: JsValue): Set[String] = {
+    println("")
+    println("JSONParse: getDatasets: " + json.toString())
```
JavierJia: I assume these `println`s for the debug purpose only?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
```diff
@@ -189,8 +189,9 @@ object QueryResolver {
   }
 
   private def resolveField(name: String, fieldMap: Map[String, Field]): Field = {
-    val field = fieldMap.get(name)
+    val field = fieldMap.get(name.split("\\.", 2).head)
     field match {
+      case Some(f) if (f.dataType == DataType.Json && name.contains(".")) => new JsonField(name, true)
```
JavierJia: is this a special rule for `MySQL`?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+import play.api.libs.json.Json
+import java.io.InputStream
+import java.util.Date
+
+
+
+/**
+  * Created by sicongliu on 17/5/19.
```
JavierJia: remove the `created` comments.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,130 @@
...
+        var rsJson: JsObject = Json.obj()
+        while (index <= columnCount) {
+          val column = rsmd.getColumnLabel(index)
+          val columnLabel = column.toLowerCase()
+          val value = result.getObject(column)
+          if (value == null) {
+            rsJson = rsJson ++ Json.obj(
+              columnLabel -> JsNull
+            )
+          } else if (value.isInstanceOf[Integer]) {
```
JavierJia: for the type match test, scala has a better syntax of using `match`:
```
value match {
  case str: String => rsJson += Json.obj( "key" -> JsString(str))
  case int : Int => ...
  case bool: Boolean => ...
}
```

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -100,7 +103,7 @@ case class TimeField(override val name: String, override val isOptional: Boolean
 
 }
 
-object TimeField {
+object TimeField {  // TODO: DateTime format might be different
```
JavierJia: any idea what's the difference?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -128,6 +131,12 @@ case class BooleanField(override val name: String, override val isOptional: Bool
 
 }
 
+case class JsonField(override val name: String, override val isOptional: Boolean = false)
```
JavierJia: mind to add some comment about why we need this new field?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -21,20 +22,27 @@ object TestQuery {
 
 
   val createAt = twitterField("create_at").asInstanceOf[TimeField]
-  val hashtags = twitterField("hashtags")
+  val hashtags = twitterField("hashtag")  //New: hashtag
```
JavierJia: why do we need to change this? if so, the existing test should fail, right?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TwitterDataStoreForSQL.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+  * Created by yamamuraisao on 17/7/20.
+  */
+
+import edu.uci.ics.cloudberry.zion.model.schema._
+
+object TwitterDataStoreForSQL{
+  val DatasetName = "twitter_ds_tweet"
+  val TimeFieldName = "create_at"
+  val TwitterSchemaForSQL = Schema("twitter.typeTweet",
```
JavierJia: is it the correct schema name in MySQL?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/DataSetInfo.scala
```diff
@@ -108,6 +108,8 @@ object DataSetInfo {
           JsSuccess(StringField(name, isOptional))
         case DataType.Time =>
           JsSuccess(TimeField(name, isOptional))
+        case DataType.Json =>
+          JsSuccess(JsonField(name, isOptional))
```
hh1680651: Hierarchy and Bag fields are stored as Json format in MySQL. So I created JsonField and DataType.Json.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
```diff
@@ -192,7 +192,10 @@ object QueryResolver {
     val field = fieldMap.get(name)
     field match {
       case Some(f) => f
-      case _ => throw FieldNotFound(name)
+      case _ => fieldMap.get(name.split("\\.", 2).head) match {
+          case Some(f) if (f.dataType == DataType.Json && name.contains(".")) => new JsonField(name, false) // name has the format of 'fieldName.subFieldName'
```
hh1680651: A request's fieldName might not be in the fieldMap, e.g.: 
Field: {
      "name" : "user.id",
      "isOptional" : false,
      "datatype" : "Json"
 }, but "user" is the keyword in fieldMap, so we simply create a new JsonField, with name equals to "user.id".

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryValidator.scala
```diff
@@ -81,7 +83,7 @@ object QueryValidator {
   }
 
   def validateUnnest(unnest: UnnestStatement): Unit = {
-    require(unnest.field.isInstanceOf[BagField], "unnest can only apply on Bag type")
+    require(unnest.field.isInstanceOf[BagField] || unnest.field.isInstanceOf[JsonField], "unnest can only apply on Bag type")
```
hh1680651: Bag type is stored as Json format in MySQL.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+    if (!result.isBeforeFirst()) {
+      Future(true)
+    }
+    try {
+      while (result.next) {
+        var index = 1
+        var rsJson: JsObject = Json.obj()
+        while (index <= columnCount) {
+          val columnLabel = rsmd.getColumnLabel(index)
+          val value = result.getObject(columnLabel)
```
hh1680651: This Try Catch is not neat.. But I don't know how to break a loop in scala when value = result.getObject(columnLabel) returns null.
Any suggestions?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+              rsJson = rsJson ++ Json.obj(columnLabel -> Json.toJson(arr.asInstanceOf[Array[String]]))
+            case str: String =>
+              try {
+                rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(str.toInt))
+              } catch {
+                case other => try {
+                  rsJson = rsJson ++ Json.obj(columnLabel -> Json.parse(str))
+                } catch {
+                  case s => rsJson = rsJson ++ Json.obj(columnLabel -> JsString(str.asInstanceOf[String]))
+                }
```
hh1680651: A string can be Json, or Number, or String. 
e.g.: "geo_tag.cityID" is stored as a subField with Json format, when selecting geo_tag->"$.cityID", the return type would be a String.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  def postQuery(query: String): Future[JsValue] = {
+    val result = statement.executeQuery(query)
+    val rsmd = result.getMetaData
+    val columnCount = rsmd.getColumnCount
+    var qJsonArray: JsArray = Json.arr()
+    if (!result.isBeforeFirst()) {
+      Future(true)
+    }
```
hh1680651: ResultSet is empty.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+          }
+          index += 1
+        }
+        qJsonArray = qJsonArray :+ rsJson
+      }
+    } catch {
+      case e => Future(true)
+    }
+    if (qJsonArray == JsArray() && query.contains("berry.meta")) {
+      Future(berry)
```
hh1680651: Table berry.meta is empty when starting neo server for the first time.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+      case q: UpsertRecord => parseUpsert(q)
+      case q: DropView => parseDrop(q, schemaMap)
+      case q: DeleteRecord => parseDelete(q, schemaMap)
+      case _ => ???
+    }
+    s"$result"
+  }
+
+  def splitSchemaMap(schemaMap: Map[String, AbstractSchema]): (Map[String, Schema], Map[String, LookupSchema]) = {
+    val temporalSchemaMap = scala.collection.mutable.Map[String, Schema]()
```
hh1680651: Copied this function from SQLPPGenerator for the parseCreate function.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+    val ddl: String = genDDL(create.dataset, sourceSchema)
+    val insert =
+      s"""
+     |replace into $quote${create.dataset}$quote ()
+     |(
+     |${parseQuery(create.query, schemaMap)}
+     |)
+      """.stripMargin.replace("T", " ").replace("Z", "")
+    ddl + insert
+  }
```
hh1680651: create table if not exist & "upsert" the query results.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+      case DataType.Hierarchy => ???
+      case DataType.Record => ???
+    }
+  }
+
+  def parseAppend(append: AppendView, schemaMap: Map[String, AbstractSchema]): String = {
+    s"""
+       |replace into $quote${append.dataset}$quote (
+       |${parseQuery(append.query, schemaMap)}
+       |)""".stripMargin.replace("T", " ").replace("Z", "")
```
hh1680651: The Date type's format only differs when inserting records: 
"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'" vs "yyyy-MM-dd HH:mm:ss.SSS"

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+    val x = q.records.value
+    var queryResult = ArrayBuffer.empty[String]
+    x.foreach {
+      record =>
+        val name: String = (record \ "name").as[JsString].value
+        val schema: JsValue = (record \ "schema").as[JsValue]
+        val dataInterval: JsValue = (record \ "dataInterval").as[JsValue]
+        val stats: JsValue = (record \ "stats").as[JsValue]
+        queryResult += (s"('${name}','${schema}','${dataInterval}','${stats}')")
+    }
```
hh1680651: Since parseUpsert() is only called when inserting schemas into berry.meta, and the fields are fixed.
Do I have to generalize this function?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+      ???
+    }
+  }
+
+  protected def initExprMap(dataset: String, schemaMap: Map[String, AbstractSchema]): Map[String, FieldExpr] = {
+    val schema = schemaMap(dataset)
+    schema.fieldMap.mapValues { f =>
+      f.dataType match {
+        case DataType.Json if f.name.contains(".") =>
+            FieldExpr(s"$sourceVar.$quote${f.name}$quote", s"$sourceVar.${f.name.split("\\.", 2).head + "->\"$." + f.name.split("\\.", 2)(1) + '"'}")
```
hh1680651: Change the defExpr when calling Json's subfield. 
e.g.: Convert user.id to user->"$.id"
I'm considering adding a function only for Json's field's transform, so this line can stop repeating.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+        case _ if (dataset.equals("berry.meta") && f.name.contains(".")) =>  //TODO: generalize berry.metaType
+          FieldExpr(s"$sourceVar.$quote${f.name}$quote", s"$sourceVar.${f.name.split("\\.", 2).head + "->\"$." + f.name.split("\\.", 2)(1) + '"'}")
```
hh1680651: TODO: solve the cases like TimeField(t.user.create_at).

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+    val lookupStr = lookups.zipWithIndex.map {
+      case (lookup, id) =>
+        val lookupExpr = s"$actualLookupVar$id"
+        val conditions = lookup.lookupKeys.zip(lookup.sourceKeys).map {
+          case (lookupKey, sourceKey) =>
+            val sourceExpr = try {
+              exprMap(sourceKey.name)
+            } catch {
+              case e => exprMap(sourceKey.name.split("\\.",2)(0))
+            }
```
hh1680651: Subfield of Json. So only it's head can be found in exprMap.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+    }
+  }
+
+  private def parseFilter(filters: Seq[FilterStatement], exprMap: Map[String, FieldExpr], unnestTestStrs: Seq[String], queryBuilder: StringBuilder): ParsedResult = {
+    if (filters.isEmpty && unnestTestStrs.isEmpty) {
+      ParsedResult(Seq.empty, exprMap)
+    } else {
+      val filterStrs = filters.map { filter =>
+        filter.field.dataType match {
+          case js: DataType.Json.type if (filter.field.name.contains(".")) =>
```
hh1680651: Calling subfield of Json, I'm considering adding a function only for Json's field's transform, so this line can stop repeating.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+    val unnestTestStrs = new ListBuffer[String]
+    val unnestStr = unnest.zipWithIndex.map {
+      case (unnest, id) =>
+        val expr = exprMap(unnest.field.name)
+        val newExpr = s"${quote}hashtags$quote"
+        producedExprs += (unnest.as.name -> FieldExpr(newExpr, newExpr))
+        if (unnest.field.isOptional) {
+          unnestTestStrs += s"${expr.refExpr} is not null"
+        }
+        s"" // TODO: literal view is not allowed in MySQL
```
hh1680651: TODO: Lateral view is not allowed in MySQL. I haven't think out corresponding syntax in MySQL. To fulfill this function (e.g.: "select hashtag, count(*) ...") we have to join the result list and return it's count. 
Any suggestions?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
hh1680651: All hierarchy & bag fields are stored as Json field in MySQL.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
```diff
@@ -0,0 +1,757 @@
...
+      val group = GroupStatement(Seq(byHashTag), Seq(aggrCount))
+      val query = new Query(TwitterDataSetForSQL, Seq.empty, Seq.empty, filter, Seq(unnestHashTag), Some(group), Some(selectTop10byHashTag))
+      val result = parser.generate(query, Map(TwitterDataSetForSQL -> twitterSchemaForSQL))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """|select t.`hashtags` as `hashtags`,count(*) as `count`
+           |from `twitter_ds_tweet` t
+           |where t.`hashtags` is not null and t.`create_at` >= '2016-01-01T00:00:00.000Z' and t.`create_at` < '2016-12-01T00:00:00.000Z' and lower(t.`text`) like '%zika%' and lower(t.`text`) like '%virus%' and t.geo_tag->"$.stateID" in ( 37,51,24,11,10,34,42,9,44 )
+           |group by `hashtags`
+           |order by count(*) desc
+           |limit 10""".stripMargin.trim)
```
hh1680651: TODO: MySQL doesn't support lateral view, because there's no Array type. I haven't found any corresponding MySQL syntax yet.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -4,20 +4,25 @@ import edu.uci.ics.cloudberry.zion.model.schema.{GeoCellThousandth, _}
 import org.joda.time.{DateTime, DateTimeZone}
 import play.api.libs.json.Json
 
+
 object TestQuery {
 
   DateTimeZone.setDefault(DateTimeZone.UTC)
   val TwitterDataSet = TwitterDataStore.DatasetName
+  val TwitterDataSetForSQL = TwitterDataStoreForSQL.DatasetName
```
hh1680651: I've tried to write a trait/class and two extends, but the code seems ugly(a lot of "lazy"s), so added these val for SQLGeneratorTest instead.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -4,20 +4,25 @@ import edu.uci.ics.cloudberry.zion.model.schema.{GeoCellThousandth, _}
 import org.joda.time.{DateTime, DateTimeZone}
 import play.api.libs.json.Json
 
+
```
hh1680651: I'll delete this in next commit.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -1052,8 +1081,10 @@ object TestQuery {
...
   def twitterField(field: String): Field = twitterSchema.fieldMap(field)
 
+  def twitterFieldForSQL(field: String): Field = twitterSchemaForSQL.fieldMap(field)
+
   def populationField(field: String): Field = populationSchema.fieldMap(field)
 
   def literacyField(field: String): Field = literacySchema.fieldMap(field)
 
-}
+}
```
hh1680651: I'll delete this in next commit.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
JavierJia: then you should update the PR?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -4,20 +4,25 @@ import edu.uci.ics.cloudberry.zion.model.schema.{GeoCellThousandth, _}
...
   val twitterSchema = TwitterDataStore.TwitterSchema
+  val twitterSchemaForSQL = TwitterDataStoreForSQL.TwitterSchemaForSQL
   val populationSchema = PopulationDataStore.PopulationSchema
   val literacySchema = LiteracyDataStore.LiteracySchema
   val startTime = "2016-01-01T00:00:00.000Z"
   val endTime = "2016-12-01T00:00:00.000Z"
   val twitterSchemaMap = Map(TwitterDataSet -> twitterSchema)
+  val twitterSchemaMapForSQL = Map(TwitterDataSetForSQL -> twitterSchemaForSQL)
   val allSchemaMap = Map(TwitterDataSet -> twitterSchema, PopulationDataSet -> populationSchema, literacyDataSet -> literacySchema)
+  val allSchemaMapForSQL = Map(TwitterDataSetForSQL -> twitterSchemaForSQL)
```
JavierJia: Can you organize the MySQL related code together at least?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,654 @@
...
+  protected def fieldType2ADMType(field: Field): String = {
+    field.dataType match {
+      case DataType.Number => "double"
+      case DataType.Time => "datetime"
+      case DataType.Point => "point"
+      case DataType.Boolean => "tinyint"
+      case DataType.String => "varchar(255)"
+      case DataType.Text => "text"
+      case DataType.Json => "json"
+      case DataType.Bag => s"{{${fieldType2ADMType(Field("", field.asInstanceOf[BagField].innerType))}}}"
```
JavierJia: we can leave the `Bag` as `???` for now.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
JavierJia: OK for this PR.

cloudberry/project/dependencies.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
hh1680651: All prints are debug only, I'll delete them when requesting code review.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryResolver.scala
hh1680651: These changes for Json field are removed.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
hh1680651: Only differs in insert queries.

cloudberry/neo/app/db/Migration_20160814.scala
```diff
@@ -11,16 +11,31 @@ private[db] class Migration_20160814() {
...
+             |create table if not exists `berry.meta` (
+             |`name` varchar(255) not null,
+             |`schema` json default null,
+             |`dataInterval` json default null,
+             |`stats` json default null,
+             |primary key(`name`)
+             |)
+             |""".stripMargin
+        }
+      case _ =>
```
hh1680651: Ok.

cloudberry/neo/conf/application.conf
```diff
@@ -88,8 +88,9 @@ berry.query.gap = "1 day"
 
 #asterixdb.url = "http://localhost:19002/aql"
 asterixdb.url = "http://localhost:19002/query/service"
-
-#acceptable values: "AQL" or "SQLPP" or "sparksql"
-asterixdb.lang = SQLPP
+mysqldb.url = "jdbc:mysql://localhost:3306/twitter_ds_tweet"
```
hh1680651: Twitter_ds_tweet is my local database name, for better understanding, I changed it to "sql".

cloudberry/project/dependencies.scala
```diff
@@ -26,6 +26,8 @@ object Dependencies {
   val zionDependencies: Seq[ModuleID] = Seq(
     "joda-time" % "joda-time" % "2.9.3",
     "com.typesafe.akka" %% "akka-actor" % "2.4.4",
+    "org.scalikejdbc" %% "scalikejdbc" % "2.5.2",
```
hh1680651: Removed.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import java.sql.{Connection, DriverManager}
+import java.io.InputStream
+import java.util.Date
+import java.text.SimpleDateFormat
+
+class SQLConn(url: String)(implicit ec: ExecutionContext) extends IDataConn {
+  val driver = "com.mysql.jdbc.Driver"
+  val username = "root"
+  val password = "2048"
```
hh1680651: Should I add 2 vals in application.conf?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+  def post(query: String): Future[WSResponse] = {
+    throw new UnsupportedOperationException
+  }
+
+  def postQuery(query: String): Future[JsValue] = {
+    val result = statement.executeQuery(query)
+    val rsmd = result.getMetaData
+    val columnCount = rsmd.getColumnCount
+    var qJsonArray: JsArray = Json.arr()
+    if (!result.isBeforeFirst()) {
```
hh1680651: Empty query result. Removed this and solved with try catch instead.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
```diff
@@ -16,6 +16,7 @@ object DataType extends Enumeration {
   val Bag = Value("Bag")
   val Hierarchy = Value("Hierarchy")
   val Record = Value("Record")
+  val Json = Value("Json")
```
hh1680651: Remove all.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Functions.scala
```diff
@@ -119,6 +119,7 @@ case class Interval(unit: TimeUnit.Value, x: Int = 1) extends GroupFunc with Sca
       case Week => x * 60 * 60 * 24 * 7
       case Month => x * 60 * 60 * 24 * 30
       case Year => x * 60 * 60 * 24 * 365
+      case Date => x * 60 * 60 * 24  // MySQL: date function would return dateTime object
```
hh1680651: Function "day" would only return "dd" format, it's special in MySQL.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

hh1680651:
Flattened structure in MySQL, deleted "Json" datatype, changed SQLGenerator and passed test cases.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+        while (index <= columnCount) {
+          val columnLabel = rsmd.getColumnLabel(index)
+          val value = result.getObject(columnLabel)
+          value match {
+            case int: Integer =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(int.asInstanceOf[Int]))
+            case boolean: java.lang.Boolean =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsBoolean(boolean.asInstanceOf[Boolean]))
+            case date: Date =>
+              val minuteFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
```
JavierJia: we have the format in `TimeField.TimeFormat`. it can be used globally.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsString(minuteFormat.format(date.asInstanceOf[Date].getTime)))
+            case long: java.lang.Long =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(long.asInstanceOf[Long]))
```
JavierJia: ditto

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+            case double: java.lang.Double =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(double.asInstanceOf[Double]))
```
JavierJia: ditto

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+            case float: java.lang.Float =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(float.asInstanceOf[BigDecimal]))
```
JavierJia: ditto

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+            case arr: Array[String] =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> Json.toJson(arr.asInstanceOf[Array[String]]))
+            case str: String =>
+              try {
+                rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(str.toInt))
```
JavierJia: why do you want to convert it to `Int`?  the middleware should not change the returned type.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+              } catch {
+                case json => try {
+                  rsJson = rsJson ++ Json.obj(columnLabel -> Json.parse(str))
+                } catch {
+                  case string =>
+                    rsJson = rsJson ++ Json.obj(columnLabel -> JsString(str.asInstanceOf[String]))
+                }
+              }
+            case any: AnyRef =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNull)
```
JavierJia: what could be the other cases? it shouldn't be `null` if it is a meaningful value.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+        }
+        qJsonArray = qJsonArray :+ rsJson
+      }
+    } catch {
+      case e =>
+        Future(defaultQueryResponse)  //TODO: check the return type
+    }
+    if (qJsonArray == JsArray() && query.contains("berry.meta")) {
+//      Future(Json.toJson(berry))  //TODO: check correctness, default is not usable
+      Future(defaultQueryResponse)
```
JavierJia: so we haven't implemented the meta data lookup, right?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      println("not empty: result is: " + qJsonArray.toString())
```
JavierJia: you can use `Logger.error` of `play.api` package instead of using `print`.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Functions.scala
JavierJia: but it is exactly what we want right? this unit is used to quantify a time range, e.g., how many days, month.... we can't not see how many date, right?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Functions.scala
```diff
@@ -132,6 +133,7 @@ object TimeUnit extends Enumeration {
   val Week = Value("week")
   val Month = Value("month")
   val Year = Value("year")
+  val Date = Value("date")
```
JavierJia: If you want to add a `Date` function for MySQL, that is totally fine. but it is not a `unit` for sure.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -124,6 +124,17 @@ object TestQuery {
...
+  val twitterSchemaForSQL = TwitterDataStoreForSQL.TwitterSchemaForSQL
+  val twitterSchemaMapForSQL = Map(TwitterDataSetForSQL -> twitterSchemaForSQL)
+  val allSchemaMapForSQL = Map(TwitterDataSetForSQL -> twitterSchemaForSQL)
+  val byGeoState = ByStatement(geoStateID, None, Some(Field.as(geoStateID, "state")))
+  val zikaCreateQueryForSQL = Query(TwitterDataSetForSQL, filter = Seq(zikaFilter))
+  val groupPopulationSumForSQL = GroupStatement(
+    bys = Seq(byGeoState),
+    aggregates = Seq(AggregateStatement(population, Sum, Field.as(Sum(population), "sum")))
+  )
+
```
JavierJia: much organized now 😄

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -124,6 +124,17 @@ object TestQuery {
     aggregates = Seq(AggregateStatement(population, Sum, Field.as(Sum(population), "sum")))
   )
 
+  val TwitterDataSetForSQL = TwitterDataStoreForSQL.DatasetName
```
JavierJia: `TwitterDataSetForSQL` --> `twitterDataSetForSQL` to be consistent.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
```diff
@@ -1052,6 +1063,8 @@ object TestQuery {
 
   def twitterField(field: String): Field = twitterSchema.fieldMap(field)
 
+  def twitterFieldForSQL(field: String): Field = twitterSchemaForSQL.fieldMap(field)
```
JavierJia: missed this one?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TwitterDataStoreForSQL.scala
```diff
@@ -0,0 +1,48 @@
...
+                                        TextField("user_mentions"),
+                                        TextField("lang"),
+                                        NumberField("id"),
+                                        NumberField("is_retweet"),
+                                        NumberField("geo_tag.stateID"),
+                                        NumberField("geo_tag.countyID"),
+                                        NumberField("geo_tag.cityID"),
+                                        StringField("geo_tag.stateName"),
+                                        StringField("geo_tag.countyName"),
+                                        StringField("geo_tag.cityName")
```
JavierJia: just curious, can the `.` be part of the field name?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
```diff
@@ -0,0 +1,713 @@
...
+      val result = parser.generate(query, Map(TwitterDataSetForSQL -> twitterSchemaForSQL))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |select `hashtags` as `tag`,count(*) as `count`
+          |from(
+          |select json_extract(`hashtags`, concat('$[', idx, ']')) as hashtags
+          |from(
+          | select t.`hashtags` as `hashtags`
+          |from `twitter_ds_tweet` t
+          |where t.`hashtags` is not null and lower(t.`text`) like '%zika%' and lower(t.`text`) like '%virus%' and t.`create_at` >= '2016-01-01 00:00:00.000' and t.`create_at` < '2016-12-01 00:00:00.000' and t.`geo_tag.stateID` in ( 37,51,24,11,10,34,42,9,44 )
```
JavierJia: just want to make sure `like '%zika%' ` will look at the index right?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TwitterDataStoreForSQL.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import edu.uci.ics.cloudberry.zion.model.schema._
+
+object TwitterDataStoreForSQL {
+  val DatasetName = "twitter_ds_tweet"
+  val TimeFieldName = "create_at"
+  val TwitterSchemaForSQL = Schema("twitter_ds_tweet",
+                                      Seq(
+                                        TimeField("create_at"),
+                                        TextField("hashtags"),
```
JavierJia: in cloudberry, the `TextField` often indicate there is an inverted index built on. I think the cloudberry type of the schema should be the same as the AsterixDB one. We only need to change the field name.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,675 @@
...
+
+  protected def genDDL(name: String, schema: Schema): String = {
+    def mkNestDDL(names: List[String], typeStr: String): String = {
+      names match {
+        case List(e) => s"  $e $typeStr"
+      }
+    }
+
+    val fields = schema.fieldMap.values.filter(f => f.dataType != DataType.Hierarchy && f != AllField).map {
+      f => mkNestDDL(f.name.split("\\.").toList, fieldType2ADMType(f) + (if (f.isOptional) " default null" else " not null"))
```
JavierJia: can we just further remove the `.` and related handling logic in this PR?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
JavierJia: it seems we should have a separate `TimeFormat` for MySQL.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,675 @@
...
+
+  def parseAppend(append: AppendView, schemaMap: Map[String, AbstractSchema]): String = {
+    s"""
+       |replace into $quote${append.dataset}$quote (
+       |${parseQuery(append.query, schemaMap)}
+       |)""".stripMargin.replace("T", " ").replace("Z", "")
+  }
+
+  def parseUpsert(q: UpsertRecord): String = {  //TODO: generalization
+    val x = q.records.value
```
JavierJia: x is a bad name..

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,675 @@
...
+    val actualLookupVar = if (inGroup) groupedLookupVar else lookupVar
+    val lookupStr = lookups.zipWithIndex.map {
+      case (lookup, id) =>
+        val lookupExpr = s"$actualLookupVar$id"
+        val conditions = lookup.lookupKeys.zip(lookup.sourceKeys).map {
+          case (lookupKey, sourceKey) =>
+            val sourceExpr = try {
+              exprMap(sourceKey.name)
+            } catch {
+              case e => exprMap(sourceKey.name.split("\\.",2)(0))
```
JavierJia: again, could we remove `.` handling logic in this PR?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,675 @@
...
+    filter.relation match {
+      case Relation.matches => {
+        val values = filter.values.map(_.asInstanceOf[String])
+        s"""$fieldExpr="${values(0)}""""
+      }
+      case Relation.!= => {
+        val values = filter.values.map(_.asInstanceOf[String])
+        s"""$fieldExpr!="${values(0)}""""
+      }
+      case Relation.contains => ???
```
JavierJia: can you fill this `???`?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,675 @@
...
+        if (filter.values.size != 2) throw new QueryParsingException(s"relation: ${filter.relation} require two parameters")
+        s"$fieldExpr >= ${filter.values(0)} and $fieldExpr < ${filter.values(1)}"
+      case Relation.in =>
+        s"$fieldExpr in ( ${filter.values.mkString(",")} )"
+      case _ =>
+        s"$fieldExpr ${filter.relation} ${filter.values(0)}"
+    }
+  }
+
+  protected def parseJsonRelation(filter: FilterStatement, fieldExpr: String): String = {
```
JavierJia: we should not have `JSON` relation in this PR?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TwitterDataStoreForSQL.scala
hh1680651: Yes, every field is quoted by "`"

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Functions.scala
hh1680651: Deleted this unit.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+import java.sql.{Connection, DriverManager}
+import java.io.InputStream
+import java.util.Date
+import java.text.SimpleDateFormat
+
+class SQLConn(url: String)(implicit ec: ExecutionContext) extends IDataConn {
+  val driver = "com.mysql.jdbc.Driver"
+  val username = "root"  //TODO: how to config?
+  val password = "2048"
+  var connection:Connection = _
```
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import java.sql.{Connection, DriverManager}
+import java.io.InputStream
+import java.util.Date
+import java.text.SimpleDateFormat
+
+class SQLConn(url: String)(implicit ec: ExecutionContext) extends IDataConn {
+  val driver = "com.mysql.jdbc.Driver"
+  val username = "root"  //TODO: how to config?
+  val password = "2048"
```
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+  var connection:Connection = _
+  Class.forName(driver)
+  connection = DriverManager.getConnection(url, username, password)
+  val stream: InputStream = getClass.getResourceAsStream("/ddl/berry.json")
```
hh1680651: Maybe we still do? When starting up and berry.meta is empty, we should return a schema first? I've tried other ways but failed.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+    try {  //TODO
+      while (result.next) {
+        var index = 1
+        var rsJson: JsObject = Json.obj()
+        while (index <= columnCount) {
+          val columnLabel = rsmd.getColumnLabel(index)
+          val value = result.getObject(columnLabel)
+          value match {
+            case int: Integer =>
+              rsJson = rsJson ++ Json.obj(columnLabel -> JsNumber(int.asInstanceOf[Int]))
```
hh1680651: JsNumber only takes BigDecimal type, so I've made some changes like int.toInt.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Removed. This case only happens in Json types in MySQL.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Added a break.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Updated.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/TestQuery.scala
hh1680651: I thought the TwitterDataSet was initialed in capitals, should I change it here?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
hh1680651: You mean I should build a index on `text` firstly?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

hh1680651:
The above commits have:
- Connected to MySQL backend.
- Support twittermap's function.
- Support metadata's update & insertion.
- Support new "view"s creation.
- Passed test cases.

cloudberry/neo/app/db/Migration_20160814.scala
```diff
@@ -11,16 +11,32 @@ private[db] class Migration_20160814() {
...
-         | name : string,
-         | stats : { createTime: string}
-         |}
-         |
-         |create dataset $berryMeta(berry.metaType) if not exists primary key name;
+    conn match {
+      case sql: SQLConn =>
+        conn.postControl {
+          s"""
+             |drop table if exists  `berry.meta`;
```
hh1680651: Removed, sorry about these testing codes!

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+  val source = scala.io.Source.fromInputStream(stream).getLines.mkString
+  val berry: JsValue = Json.parse(source)
+  val defaultQueryResponse = Json.toJson(Seq(Seq.empty[JsValue]))
+  val statement = connection.createStatement
+  def post(query: String): Future[WSResponse] = {
+    throw new UnsupportedOperationException
+  }
+
+  def postQuery(query: String): Future[JsValue] = {
+    println("postQuery: " + query)
```
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,100 @@
...
+    val result = statement.executeQuery(query)
+    val rsmd = result.getMetaData
```
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Not implemented yet.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Changed into records.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/AbstractSchema.scala
JavierJia: please be specific in the comment.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
JavierJia: yes.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
JavierJia: the meta will be empty at first. then the code in `twittermap/web/model/model/Migration_20170428.scala` will register the request dataset.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,97 @@
...
+
+class SQLConn(url: String)(implicit ec: ExecutionContext) extends IDataConn {
+  val driver = "com.mysql.jdbc.Driver"
+  val username = "root"
+  val password = "2048"
+  var connection:Connection = _
+  Class.forName(driver)
+  connection = DriverManager.getConnection(url, username, password)
+  val defaultQueryResponse = Json.toJson(Seq(Seq.empty[JsValue]))
+  val stream : InputStream = getClass.getResourceAsStream("/ddl/berry.json")
```
hh1680651: Removed.

cloudberry/neo/app/db/Migration_20160814.scala
```diff
@@ -11,16 +11,32 @@ private[db] class Migration_20160814() {
...
+             |create table if not exists `berry.meta` (
+             |`name` varchar(255) not null,
+             |`schema` json default null,
```
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
hh1680651: Updated.

cloudberry/neo/conf/application.conf
```diff
@@ -88,8 +88,9 @@ berry.query.gap = "1 day"
 
 #asterixdb.url = "http://localhost:19002/aql"
 asterixdb.url = "http://localhost:19002/query/service"
-
-#acceptable values: "AQL" or "SQLPP" or "sparksql"
-asterixdb.lang = SQLPP
+mysqldb.url = "jdbc:mysql://localhost:3306/sql?user=root&passwd=2048"
```
waans11: Is there a way not to specify plain account and password here?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,618 @@
...
+    */
+  case class ParsedResult(strs: Seq[String], exprMap: Map[String, FieldExpr])
+
+
+  protected val sourceVar: String = "t"
+
+  protected val jsonVar: String = "$"
+
+  protected val appendVar: String = "ta"
+
```
waans11: Maybe remove empty lines?

cloudberry/neo/conf/application.conf
hh1680651: I haven't found any solution to it.. maybe user have to change mysql's config file..

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,104 @@
...
+  def post(query: String): Future[WSResponse] = {
+    throw new UnsupportedOperationException
+  }
+
+  def postQuery(query: String): Future[JsValue] = query match {
+    case berry if query.contains("`berry.meta`") => postBerryQuery(query)
+    case _ => postGeneralQuery(query)
+  }
+
+  def postGeneralQuery(query: String): Future[JsValue] = {
```
JavierJia: make it private?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,104 @@
...
+    val resultMetadata = result.getMetaData
+    val columnCount = resultMetadata.getColumnCount
+    var qJsonArray: JsArray = Json.arr()
+    while (result.next) {
+      var columnId = 1
+      var rsJson: JsObject = Json.obj()
+      while (columnId <= columnCount) {
+        val columnLabel = resultMetadata.getColumnLabel(columnId)
+        val value = result.getObject(columnLabel)
+        breakable {
```
JavierJia: I think the `breakable` should before line 45.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
```diff
@@ -0,0 +1,104 @@
...
+          }
+        }
+        columnId += 1
+      }
+      qJsonArray = qJsonArray :+ rsJson
+    }
+    Future(Json.toJson(qJsonArray))
+  }
+
+  def postBerryQuery(query: String): Future[JsValue] = {
```
JavierJia: private ?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,606 @@
...
+        ParsedResult(Seq.empty, exprMap)
+    }
+  }
+
+  private def parseProject(exprMap: Map[String, FieldExpr]): String = {
+    exprMap.filter {
+      case (field, expr) => field != "*" && expr.refExpr != sourceVar
+    }.map {
+      case (field, expr) =>
+        if (s"${expr.defExpr}" == s"$quote$second$quote" || s"${expr.defExpr}" == s"$quote$minute$quote" || s"${expr.defExpr}" == s"$quote$hour$quote" || s"${expr.defExpr}" == s"$quote$day$quote" || s"${expr.defExpr}" == s"$quote$month$quote" || s"${expr.defExpr}" == s"$quote$year$quote"){
```
JavierJia: why is this part so complicated? and I even see the `create_at` appears. it should be general purpose generator.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
```diff
@@ -0,0 +1,786 @@
...
+        """
+          |select t.`place.full_name` as `place.full_name`,t.`place.bounding_box` as `place.bounding_box`,t.`place.country_code` as `place.country_code`,t.`user.friends_count` as `user.friends_count`,t.`user.description` as `user.description`,t.`favorite_count` as `favorite_count`,t.`geo_tag.countyID` as `geo_tag.countyID`,t.`user.location` as `user.location`,t.`place.type` as `place.type`,t.`geo_tag.cityName` as `geo_tag.cityName`,t.`user.id` as `user.id`,t.`geo_tag.stateName` as `geo_tag.stateName`,t.`geo_tag.cityID` as `geo_tag.cityID`,t.`is_retweet` as `is_retweet`,t.`text` as `text`,t.`user.screen_name` as `user.screen_name`,t.`retweet_count` as `retweet_count`,t.`place.country` as `place.country`,t.`in_reply_to_user` as `in_reply_to_user`,t.`user.statues_count` as `user.statues_count`,t.`id` as `id`,t.`coordinate` as `coordinate`,t.`place.id` as `place.id`,t.`in_reply_to_status` as `in_reply_to_status`,t.`geo_tag.stateID` as `geo_tag.stateID`,t.`create_at` as `create_at`,t.`user.create_at` as `user.create_at`,t.`place.name` as `place.name`,t.`lang` as `lang`,t.`user.lang` as `user.lang`,t.`user.name` as `user.name`,t.`geo_tag.countyName` as `geo_tag.countyName`
```
JavierJia: why can't we just return `t`?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
```diff
@@ -0,0 +1,786 @@
...
+
+    //28
+    "translate a text contain filter and group by geocell 10th" in {
+      val filter = Seq(textFilter)
+      val group = GroupStatement(Seq(byGeocell10), Seq(aggrCount))
+      val query = new Query(TwitterDataSetForSQL, Seq.empty, Seq.empty, filter, Seq.empty, Some(group), None)
+      val result = parser.generate(query, Map(TwitterDataSetForSQL -> twitterSchemaForSQL))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |select st_astext(Point(truncate(st_x(t.`coordinate`),1),truncate(st_y(t.`coordinate`),1)))  as `cell`,count(*) as `count`
```
JavierJia: what does this `st_astext` do? and what will the result look like?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
```diff
@@ -0,0 +1,786 @@
...
+
+    //31
+    "translate a text contain filter and group by bin" in {
+      val filter = Seq(textFilter)
+      val group = GroupStatement(Seq(byBin), Seq(aggrCount))
+      val query = new Query(TwitterDataSetForSQL, Seq.empty, Seq.empty, filter, Seq.empty, Some(group), None)
+      val result = parser.generate(query, Map(TwitterDataSetForSQL -> twitterSchemaForSQL))
+      removeEmptyLine(result) must_== unifyNewLine(
+        """
+          |select round(t.`geo_tag.stateID`/10)*10 as `state`,count(*) as `count`
```
JavierJia: what does this `round` function do? the `query` doesn't ask for it. 
I guess it's for processing the `int`? is there any other easy way to handle the int case?

hh1680651: It's actually caused by "byBin(10)" groupStatement (TestQuery.scala 108):
    val bin10 = Bin(10)
    val byBin = ByStatement(geoStateID, Some(bin10), Some(Field.as(bin10(geoStateID), "state")))
So I thought "Bin" function should be like this?

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
hh1680651: st_astext: print or return the geometric results as text, e.g.: "POINT(1, 2)",
without this function, the query would return a messed code.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
hh1680651: Thanks for remind! I've fixed this bug. 
It seems that sparkSQL also has this bug unfixed, do you want me to fix it?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
hh1680651: Corrected. This problems also happens in sparkSQL, do you want me to fix it?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLConn.scala
hh1680651: Ok.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
hh1680651: Other test cases are also simplified.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
JavierJia: please add this as a comment to the implementation of the `geoCell` function.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGeneratortest.scala
JavierJia: Oh, I see. got it. 👍

JavierJia:
Can you also take care of the codacy's suggestion as well? https://www.codacy.com/app/ISG-Cloudberry/cloudberry/pullRequest?prid=784871&/login. You don't necessarily to address all comments. some of issues make sense.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/SQLGenerator.scala
```diff
@@ -0,0 +1,601 @@
...
+        }
+      }
+      val filterStr = (unnestTestStrs ++ filterStrs).mkString("where ", " and ", "")
+      appendIfNotEmpty(queryBuilder, filterStr)
+      ParsedResult(Seq.empty, exprMap)
+    }
+  }
+
+  protected def parseGeoCell(scale: Integer, fieldExpr: String, dataType: DataType.Value): String = {
+    s"$geoAsText($dataType($truncate(${pointGetCoord(0)}($fieldExpr),$scale),$truncate(${pointGetCoord(1)}($fieldExpr),$scale))) "
```
hh1680651: MySql ST_ASTEXT:
https://dev.mysql.com/doc/refman/5.7/en/gis-format-conversion-functions.html
MySql ST_X, ST_Y:
https://dev.mysql.com/doc/refman/5.6/en/gis-point-property-functions.html
MySql truncate:
http://www.w3resource.com/mysql/mathematical-functions/mysql-truncate-function.php
The other functions are mostly the same as other DB.

JavierJia:
@chenlica I had a pass on this PR. Do you want to have a pass?

chenlica:
Let's do a F2F discussion about this PR.

JavierJia:
#372

PR #384, opened 2017-08-02
Update application.conf to increase the cap of WebSocket messages

DeyuChen:
Increase the cap of WebSocket messages

### DIFF
--- a/cloudberry/neo/conf/application.conf
+++ b/cloudberry/neo/conf/application.conf
@@ -93,3 +93,5 @@ asterixdb.url = "http://localhost:19002/query/service"
 asterixdb.lang = SQLPP
 #asterixdb.lang = sparksql
 asterixdb.view.meta.name = "viewMeta"
+
+play.websocket.buffer.limit = 5M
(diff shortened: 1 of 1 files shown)


cloudberry/neo/conf/application.conf
```diff
@@ -93,3 +93,5 @@ asterixdb.url = "http://localhost:19002/query/service"
 asterixdb.lang = SQLPP
 #asterixdb.lang = sparksql
 asterixdb.view.meta.name = "viewMeta"
+
+play.websocket.buffer.limit = 5M
```
waans11: Maybe put a comment here? Why 5M needs to be allocated?

chenlica: Agreed, especially since so much effort was put to this one-line change :-)

DeyuChen: Actually, for TwitterMap example, 5M might be a bit too much. I just put a safe value for future cases.

waans11:
LGTM

Issue #382, opened 2017-08-01
WebSocket message size limit


DeyuChen:
Cloudberry uses Play Framework's WebSocket API. By default it has a 65536 byte size limit for incoming messages. It causes the socket to close when the frontend sends a large query to the Cloudberry, e.g. when there are a lot of visible cities in TwitterMap. It would be better to extend the limit to a few MB.

PR #385, opened 2017-08-09
Deyu chen websocket

DeyuChen:
Drop geoIds filter when the length of the geoIds is longer than 2000 to avoid large queries.

### DIFF
--- a/examples/twittermap/web/public/javascripts/common/services.js
+++ b/examples/twittermap/web/public/javascripts/common/services.js
@@ -90,12 +90,8 @@ angular.module('cloudberry.common', [])
       queryStartDate.setDate(queryStartDate.getDate() - maxDay);
       queryStartDate = parameters.timeInterval.start > queryStartDate ? parameters.timeInterval.start : queryStartDate;
 
-      return [
+      var filter = [
         {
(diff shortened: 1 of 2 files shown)


examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -105,6 +101,16 @@ angular.module('cloudberry.common', [])
           values: keywords
         }
       ];
+      if (parameters.geoIds.length <= 2000){
```
waans11: Is this an estimation? Or based on the real number?

DeyuChen: The length is the real number of geoIds that will be placed in the query. If you mean the number 2000, it is just a safe estimation._I tried it on the textdb server and it won't cause parse exceptions. I don't have an exact limit because the size limit seems to depend on the machine. The same query may work on one machine but fail on another. Even when I send the same query to the same machine, if the size is in a certain range, it may both work or fail.

waans11:
LGTM

JavierJia:
@DeyuChen can you merge this one if it's already done?

JavierJia:
#382

PR #388, opened 2017-08-12
Can set the cache threshold in the config.

waans11:
- Can set the cache threshold in the config.
- Set the default cache threshold as 30,000.
- If the cache threshold is a non-positive value, the caching feature can be disabled.
  (Mainly for a temporary debug purpose. This can be removed once we are sure.)
- Two javascript error fix
  (resetGeoIDs and addLayer)


JavierJia:
If you check the travis log you can find the following compilation errror:
`/home/travis/build/ISG-ICS/cloudberry/examples/twittermap/web/app/controllers/MySqlTwitterMapApplicaton.scala:37: not enough arguments for method apply: (title: String, ws: String, startDate: String, endDate: Option[String], sentimentEnabled: Boolean, sentimentUDF: String, removeSearchBar: Boolean, predefinedKeywords: Seq[String], cacheThreshold: Option[String], isDrugMap: Boolean, isMySQL: Boolean)`

There is one more `MySqlTwitterMapApplicaton` that calls the `index.scala.html`. We should update it as well.

waans11:
@JavierJia thanks! I have uploaded a new patch.

PR #389, opened 2017-08-15
MySql quick start

hh1680651:
Brief quick start doc for Twittermap with MySqlDB: 
https://github.com/ISG-ICS/cloudberry/wiki/Setup-TwitterMap-Using-Cloudberry-and-MySqlDB


examples/twittermap/script/dockerRunMySqlDB.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Starting docker-machine ..."
+docker-machine create --driver virtualbox default
+docker-machine start
+eval $(docker-machine env)
+
+HOST="$(echo -e "$(docker-machine ip)" | sed -e 's/^[[:space:]]*//')"
+PORT=3308
+
+echo "Editing config files ..."
```
JavierJia: why do we need to edit the cloudberry config file in the script? 
I think the script should only take care of the setup the MySQL part. The config file is meant to be changed by a human.

examples/twittermap/script/dockerRunMySqlDB.sh
```diff
@@ -0,0 +1,52 @@
...
+#       CREATED: 08/14/2017 23:24:01 PM PDT
+#      REVISION:  ---
+#===============================================================================
+
+set -o nounset                      # Treat unset variables as an error
+
+echo "Starting docker-machine ..."
+docker-machine create --driver virtualbox default
+docker-machine start
+eval $(docker-machine env)
```
JavierJia: why do we need to start the docker-machine? it seems unnecessary.

hh1680651: Thanks! I've checked and removed docker-machine, using localhost's url instead.

examples/twittermap/script/dockerRunMySqlDB.sh
hh1680651: So.. Should I write these 3 command lines into set up step's instead?

JavierJia:
Yes.

PR #390, opened 2017-08-16
Fix MySql-TwitterMap

hh1680651:
Fix Twittermap controller & config files for MySqlDB.


examples/twittermap/web/conf/routes
```diff
@@ -5,7 +5,7 @@
 # TwitterMap routes
 GET        /                                    controllers.TwitterMapApplication.index
 GET        /drugmap                             controllers.TwitterMapApplication.drugmap
-GET        /mysqlTwittermap                     controllers.MySqlTwitterMapApplication.mysqlTwittermap
+GET        /mysqlTwittermap                     controllers.TwitterMapApplication.index
```
JavierJia: we should remove this path, right?

hh1680651: Yes, we can remove that.

PR #393, opened 2017-08-23
Support hashtag in general sql

hh1680651:
Support hashtag function in general sql's twitterMap: PostgreSQL & MySQL (and other sql DB) can share the same schema & front-end query.


JavierJia:
Cool. Any test cases?

PR #395, opened 2017-08-24
Adjust city zoom level + border color change

waans11:
- Adjust the city zoom level from 8 to 10
- Adjust the border colors of states, county, and city to identical 
  so that mouse out after a hover does not change the original border color


chenlica:
Thanks for the fix!  It made our demo more impressive!

Issue #391, opened 2017-08-18
Add a finish FLAG in the middleware to the client


JavierJia:
.. to notify the front-end the query is finished.

PR #397, opened 2017-08-24
Add the "done message when slicing is done

JavierJia:
This PR adds a `{key: "done"}` message to the client to notify that the slicing process is done. 
It will only be sent in slicing case.


JavierJia:
#391

PR #400, opened 2017-08-30
PostgreSQL for Twittermap

hh1680651:
Support PostgreSQL for Twittermap:
- Added Migration file for psql.
- Added service-postgresql.js for sending queries.
- Added switch cases for controller & html files.

### DIFF
--- a/examples/twittermap/web/app/controllers/TwitterMapApplication.scala
+++ b/examples/twittermap/web/app/controllers/TwitterMapApplication.scala
@@ -3,7 +3,7 @@ package controllers
 import java.io.{File, FileInputStream}
 import javax.inject.{Inject, Singleton}
 
-import model.{Migration_20170428, MySqlMigration_20170810}
+import model.{Migration_20170428, MySqlMigration_20170810, PostgreSqlMigration_20172829}
 import org.joda.time.DateTime
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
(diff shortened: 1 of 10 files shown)


examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -30,21 +30,22 @@ class TwitterMapApplication @Inject()(val wsClient: WSClient,
   val endDate : Option[String] = config.getString("endDate")
   val cities: List[JsValue] = TwitterMapApplication.loadCity(environment.getFile(USCityDataPath))
   val cacheThreshold : Option[String] = config.getString("cacheThreshold")
-  val isMySql: Boolean = config.getBoolean("isMySql").getOrElse(false)
+  val sqlDB: String = config.getString("sqlDB").getOrElse("false")
```
JavierJia: can you make it an enum instead of `string`? e.g., `DBType`?

examples/twittermap/web/public/javascripts/common/services-postgresql.js
```diff
@@ -0,0 +1,289 @@
+angular.module('cloudberry.common', [])
```
JavierJia: one more comment: this file and other `service-xx.js`s have the common query generator functions. could you write a `queryGenerator` module that takes care of the common generation logic? then the `service` model could be possibly reused. Thanks!

JavierJia:
#372

PR #401, opened 2017-08-31
Cloudberry config reconstruction

hh1680651:
Split cloudberry's configuration of SqlDB into twittermap/web's application-mysql.conf & application-postgresql.conf


hh1680651:
Now the config files' structure would be:
$ tree web/conf/
web/conf/
├── application-asterixdb.conf
├── application-mysqldb.conf
├── application-postgresqldb.conf
├── application.conf
├── logback.xml
├── production.conf
└── routes

JavierJia:
to make it specific for the cloudberry config, may be it's better to rename it as `cloudberry-asterixdb.conf`

PR #402, opened 2017-09-01
Add reconstructAdmSampleTweetToJson.sh

hh1680651:
Python code for changing .adm sample tweet into .json file.

### DIFF
--- a/examples/twittermap/script/reconstructAdmSampleTweetToJson.sh
+++ b/examples/twittermap/script/reconstructAdmSampleTweetToJson.sh
@@ -1,4 +1,21 @@
 #!/usr/bin/env python
+#===============================================================================
+#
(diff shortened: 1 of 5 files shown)


examples/twittermap/script/reconstructAdmSampleTweetToJson.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
```
JavierJia: add a usage? and it should be `ToMySQLJson` or `ToPostgreJSON`?

hh1680651: It's the same for MySQL or PostgreSQL. The zip file's name might be confusing, so I changed its name from `mysqlSample.json.zip` into `sample.json.zip`.

PR #406, opened 2017-09-14
Caching map results with incremental computation

waans11:
This work is done by @dharini-s and @vacha19.

[Issue #373]:

- Modified common module to interact with the map result cache module.
- Created map result cache module that caches map results and 
   performs an incremental computation.
- Added a new dependency library hashmap.js.

Other minor fix done by @waans11:
- Changed the current favicon from hyracks to Cloudberry.
- Changed the example search keyword from "zika" to "hurricane".


waans11:
Since this work is done and reviewed by both @waans11 and @JavierJia, I am going to merge this now. The part that is added by @waans11 is minor.

PR #415, opened 2017-10-07
Add the query slicing config option

waans11:
Add an option that can choose to conduct the query slicing or not.


waans11:
Depends on a situation, a user may want to disable it. Currently, there is no option. In addition, I think adding this option doesn't hurt anything. By default, we do the query slicing unless a user touches this option.

waans11:
Thanks!

Issue #414, opened 2017-10-06
Rearchitect the TwitterMap web server


JavierJia:
Currently, the architecture of the TwitterMap is as [this picture](http://cloudberry.ics.uci.edu/img/quick-start-architecture.png). The web server only generates the front-end UI and then the front-end JS talks with Cloudberry separately. It serves the purpose of demonstrating the usage of Cloudberry, however, the TwitterMap application is largely limited by this architecture. On one side, it is easier to implement the logic on a server than on a web page. On the other side, it is difficult for a browser to process the heavy computation. 

We propose a new architecture where the TwitterMap *server* will take care of communicating with Cloudberry, and the web page only talks with the TwitterMap server. It is a much cleaner architecture and it enables more optimization (e.g., server cache) for the TwitterMap application.

JavierJia:
@haochen07 Could you add a new picture to reflect the targeting architecture? Thanks.

haochen07:
Here is the new image.

![cloudberry_architecture](https://user-images.githubusercontent.com/19794264/31304478-91fc622c-aad6-11e7-8b3b-b62cf09a29c5.png)

chenlica:
Great diagram.  Thanks!

haochen07:
The architecture below illustrates how twitterMap will work in detail.

![twittermap_architecture](https://user-images.githubusercontent.com/19794264/31465017-8fceb1d0-ae87-11e7-842b-83ec3eda22e4.png)

We introduce an actor in TwitterMap server called TwitterMap Pigeon, which is responsible for rendering all user requests come from front end page and transmitting to cloudberry with respect to cloudberry Rest API. It handles the response from cloudberry and forwards to twitterMap front end as well.

The advantage of this change is to largely reduce the code complexity in front end's scripts and as we immigrant everything to the server, it makes twitterMap more extensible for new features and communication channels.

haochen07:
As the TwitterMap server is on, we need now to redefine the json request sent from front-end to TwitterMap server so that the front-end only transmits necessary information instead of fully formatted cloudberry requests. This design draft discuss how we define the front-end requests.

```
{
  click: {
    keyword: [],
    samples: [],
    normalization: true/false,
    ...
  } 
  drag: {
    map: {}
    timeline: {},
    ...
  }
  ...
}
```

chenlica:
[...]

PR #416, opened 2017-10-15
[Issue #414] Introduce TwitterMap Server

haochen07:
Related to issue #414 

**Note**
- Open websocket flow in twitterMap server.
- Implement new actor TwitterMap Pigeon to forward requests.
- Introduce Jetty Websocket client library (9.4.x) and transfer websocket request/response to/from cloudberry

**Test**
- Tested with TwitterMap browser.


chenlica:
That is fast!   Does TwitterMap work for this new architecture?

If it's done, I suggest @JavierJia first review it.  After that, we can have a meeting with @waans11 and @baiqiushi to go through the new architecture.  If it doesn't need much time, we can do it during our weekly meeting.

haochen07:
@chenlica Yes. The TwitterMap works in this architecture. 
The task is not finished. But this PR shows a milestone on the way where twittermap server is able to transfer websocket request and response between frontend and cloudberry.
My idea is to have someone, maybe @JavierJia , review it, to see if this is the right direction to go. And from there, we can do either 1) polish this PR, push it and make the rendering part in another PR, or 2) add rendering logic in the current PR and push everything together.

examples/twittermap/project/dependencies.scala
```diff
@@ -52,6 +52,7 @@ object Dependencies {
     "org.webjars" % "bootstrap" % "3.3.6",
     "org.webjars" % "flot" % "0.8.0",
     "org.webjars" % "angularjs" % "1.5.0",
+    "org.eclipse.jetty.websocket" % "websocket-client" % "9.4.7.v20170914",
```
JavierJia: wow, the version number suggests it is a reliable library 😁

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import akka.actor._
+import akka.stream.Materializer
+import org.eclipse.jetty.websocket.client.{ClientUpgradeRequest, WebSocketClient}
+import play.api.libs.json.JsValue
+import socket.TwitterMapClientSocket
+
+import scala.concurrent.ExecutionContext
+
+class TwitterMapPigeon (val cloudberryWS: String,
```
JavierJia: Add scala doc in front of the class?

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
```diff
@@ -0,0 +1,43 @@
...
+                        val out: ActorRef)
+                       (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
+
+  val twitterMapWebSocketClient: WebSocketClient = new WebSocketClient
+  val twitterMapClientSocket: TwitterMapClientSocket = new TwitterMapClientSocket(out)
```
JavierJia: the names of these two variables are too similar. I couldn't figure out what does each mean. can you elaborate?

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override def preStart(): Unit = {
+    super.preStart
+    twitterMapWebSocketClient.start()
+    twitterMapWebSocketClient.connect(twitterMapClientSocket, new URI(cloudberryWS), new ClientUpgradeRequest)
+    twitterMapClientSocket.getLatch.await()
```
JavierJia: what's the purpose of `getLatch` and `await`?

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
```diff
@@ -0,0 +1,43 @@
...
+                        val out: ActorRef)
+                       (implicit ec: ExecutionContext, implicit val materializer: Materializer) extends Actor with ActorLogging {
+
+  val twitterMapWebSocketClient: WebSocketClient = new WebSocketClient
+  val twitterMapClientSocket: TwitterMapClientSocket = new TwitterMapClientSocket(out)
+
+  override def preStart(): Unit = {
+    super.preStart
+    twitterMapWebSocketClient.start()
+    twitterMapWebSocketClient.connect(twitterMapClientSocket, new URI(cloudberryWS), new ClientUpgradeRequest)
```
JavierJia: what is `ClientUpgradeRequest`?

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -176,6 +189,41 @@ object TwitterMapApplication {
...
   }
 
+  /**
+    * Copy and revised based on [[play.api.libs.streams.ActorFlow.actorRef()]]
+    * The difference is that instead of using implicit '''ActorRefFactory''' to instantiate actors,
+    * we use the '''ActorSystem''' parameter that passed to the method.
+    * The reason is that we don't want the implicit '''ActorRefFactory''' here because it has been used in CloudBerry and
+    * TwitterMap application should have an independent ActorSystem.
+    */
+  def actorRef[In, Out](system: ActorSystem, props: ActorRef => Props, bufferSize: Int = 16, overflowStrategy: OverflowStrategy = OverflowStrategy.dropNew)
```
JavierJia: should it be reused from `zion`?

examples/twittermap/web/app/socket/TwitterMapClientSocket.scala
```diff
@@ -0,0 +1,50 @@
...
+import akka.actor.ActorRef
+import org.eclipse.jetty.websocket.api.Session
+import org.eclipse.jetty.websocket.api.annotations.{OnWebSocketConnect, OnWebSocketMessage, WebSocket}
+import play.api.Logger
+import play.api.libs.json.{JsValue, Json}
+
+@WebSocket
+class TwitterMapClientSocket(out: ActorRef) {
+  private var sessionOpt: Option[Session] = None
+  private val latch = new CountDownLatch(1)
```
JavierJia: what's the purpose of this `latch`?

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
haochen07: Since websocket connection begins with an upgrade request in Http protocol, `clientUpgradeRequest` serves for that. I checked the source code of websocket client that it actually can be absent from parameters, which will be internally set up. So I removed it in the updated commits.

Details here:
https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/websocket/client/WebSocketClient.html

examples/twittermap/web/app/actor/TwitterMapPigeon.scala
haochen07: The `countDownLatch` is used for concurrent control which guaranteed that the socket session is not accessed, interrupted or closed by any other thread. The connection is open until we call `close` method when the actor is killed. I saw many code samples are using it to secure the websocket connection and I thought we need it as well in our case so I put it here. It is open to discussion.

Some source here:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html
https://github.com/jetty-project/embedded-jetty-websocket-examples/pull/2/files

examples/twittermap/web/app/socket/TwitterMapClientSocket.scala
haochen07: ditto

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
haochen07: I had seen the code snippet of `actorRef` in `zion` and I didn't think they are handling the same problem. In `TwitterMap`, the pain point is that we wanna use another `actorSystem` different from implicit one and since the default `actorRef` relying on that, we have to change it. In `zion`, I am not completely understood why we implement that but seems it still uses implicit `actorRefFactory`. That is why we cannot reuse it.

JavierJia: I got it now. 
But can you check if we can directly use the system `play.api.libs.streams.ActorFlow.actorRef()` without have to write our own?

haochen07: If we want to use a different actor system other than implicit, then we cannot use the default `ActorFlow.actorRef()` since it relies on implicit `actorSystem`

waans11:
@haochen07 and @JavierJia: Thanks for a F2F meeting today. @haochen07: Most parts look good to me. I will have one more pass once @haochen07 adds a unit test case to this pull request.

haochen07:
Due to the fact that, the new actor (TwitterMapPigeon) calls the `WebSocketClient`, which applies a superclass method that cannot be overridden (designated as `final` in the library), after discussing with @JavierJia , we decided not to unit test the transmit logic as long as the TwitterMap works. Note that if further computational logic is implemented in the actor, we claim them as `static` method which will be able to do unit testing.
@waans11 @baiqiushi

chenlica:
LGTM.

haochen07:
Are there any comments on the current PR?

waans11:
I will check out this branch and test it on local by tomorrow.

haochen07:
Yeah sure. Let me know if you have any concerns.

waans11:
Tested on my machine and it worked well.

haochen07:
Good. And @baiqiushi any comments?
I don't have authority to write so if everything goes well, feel free to merge it.

chenlica:
If both @waans11 and @JavierJia approve it, please go ahead to do the merge.

PR #423, opened 2017-10-24
Remove a duplicated query submission and an unnecessary parameter.

waans11:
- Remove a duplicated query submission when dragging the current region at city level on the Twittermap.
- Remove an unnecessary parameter (queryType) in Twittermap.


chenlica:
I suggest @baiqiushi be a reviewer in addition to @JavierJia ?

PR #425, opened 2017-10-27
Fix Query Slicing request bug on the front-end

waans11:
- Fix Query Slicing request bug on the front-end (type casting issue)
- Integrate two map result cache handling case (partial hit + complete miss)
- Map result cache now returns the result if it truly contains something (add undefined handling)


chenlica:
@JavierJia : please review it so that we can merge and push it to the live system.

PR #431, opened 2017-11-09
Framework for heatmap and pointmap

DeyuChen:
The framework for adding heatmap and pointmap separately.

### DIFF
--- a/examples/twittermap/web/public/javascripts/common/services.js
+++ b/examples/twittermap/web/public/javascripts/common/services.js
@@ -41,7 +41,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     };
   })
-  .service('cloudberry', function($rootScope, $timeout, cloudberryConfig, MapResultCache) {
+  .service('cloudberry', function($rootScope, $timeout, cloudberryConfig, MapResultCache) { // use $rootScope to share information between servcies and controllers
     var startDate = config.startDate;
     var endDate = config.endDate;
     var defaultNonSamplingDayRange = 1500;
--- a/examples/twittermap/web/public/javascripts/common/services.js
+++ b/examples/twittermap/web/public/javascripts/common/services.js
@@ -41,7 +41,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     };
   })
-  .service('cloudberry', function($timeout, cloudberryConfig, MapResultCache) {
+  .service('cloudberry', function($rootScope, $timeout, cloudberryConfig, MapResultCache) {
     var startDate = config.startDate;
     var endDate = config.endDate;
     var defaultNonSamplingDayRange = 1500;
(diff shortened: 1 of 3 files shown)


examples/twittermap/web/public/javascripts/map/controllers.js
```diff
@@ -691,30 +717,44 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common','clou
...
-        }
-        if(newResult['doSentiment'] !== oldValue['doSentiment']) {
-          $scope.doSentiment = newResult['doSentiment'];
-          if($scope.doSentiment) {
-            $scope.infoPromp = "Score";  // change the info promp
-          } else {
-            $scope.infoPromp = config.mapLegend;
-          }
-          drawMap($scope.result);
+        switch ($rootScope.maptype) {
```
waans11: For this part, If the "mapResult" name will be used for all three types of map, I think "if (newResult['mapResult'] !== oldValue['mapResult']) {" should be placed first before switch. Since this watchCollection checks whether the mapResult is changed, then conducts some necessary operations. 

if (new !== old) {

switch ($rootScope.maptype) {
   case ...
}

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -41,7 +41,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     };
   })
-  .service('cloudberry', function($timeout, cloudberryConfig, MapResultCache) {
+  .service('cloudberry', function($rootScope, $timeout, cloudberryConfig, MapResultCache) {
```
waans11: Where do we create this $rootScope? Can we put some comments there? What does $rootScope mean...

DeyuChen: $rootScope is a angularJS thing. It is a way to share information between different controllers and services.

examples/twittermap/web/public/javascripts/map/controllers.js
DeyuChen: I just have concern about "doNormalization" and "doSentiment". Only countmap needs them. E.g. when we first switch from countmap to pointmap, drawCountMap will be called because new doNormalization result != old result. Also, heatmap and pointmap may require their own additional checks in the future.

examples/twittermap/web/public/javascripts/common/services.js
DeyuChen: I have added some comments about $rootScope

DeyuChen: Sorry I just found a way to avoid using $rootScope in common/services. We can already use the cloudberry.parameters to share information between different controllers. And the maptype variable can be passed to common/services this way. I will commit a new version.

baiqiushi: OK. But I can not see anything wrong with put the maptype in $rootScope.

DeyuChen: I just want to avoid passing $rootScope to common/services. Although I am not sure if there is any performance issue, but the original logic seems to try to avoid that.

chenlica:
Super!  Looking forward to the next map!

Issue #432, opened 2017-11-09
Enable "limit" feature to cloudberry queries


haochen07:
Currently, cloudberry will keep returning sliced query results till all records have been scanned. We would like to take advantage of the `limit` API, but change the internal implementation so that users could limit the number of results for slicing queries and cloudberry would be responsible for this process.

haochen07:
### Update
For batch requests, users may want to limit the number of returned results for each single requests respectively. Thus the above implementation will not work. And this requires an API change.

Currently, we have the `select` and `limit` field for single requests that limits the number of results. However, that is executed in DB level and thus has been used only for non-slicing queries such as tweets samplings. **The key part of the new design is to distinguish the two kinds of `limit` commands: one is the one we already have for sampling requests, the other is the new one we would like to implement for slicing queries.**

#### Proposal 1
Add a new field `type` to the requests to distinguish `slicing` and `nonSlicing` queries. And limit number will appear in `select` -> `limit`.

```
{
    "batch": [{
        "dataset": "twitter.ds_tweet",
        "filter": [{
            "field": "create_at",
            "relation": "inRange",
            "values": ["2016-01-01T00:00:00.000Z", "2016-12-31T00:00:00.000Z"]
        }, {
            "field": "text",
            "relation": "contains",
            "values": ["zika", "virus"]
        }],
        "group": {
            "by": [{
                "field": "geo_tag.stateID",
                "as": "state"
            }, {
                "field": "create_at",
                "apply": {
                    "name": "interval",
                    "args": {
                        "unit": "hour"
                    }
                },
                "as": "hour"
            }],
            "aggregate": [{
                "field": "*",
                "apply": {
                    "name": "count"
                },
                "as": "count"
            }],
            "select": {
                "limit": 1000,
            },
            "type": "slicing"
        }
    }, {
        "dataset": "twitter.ds_tweet",
        "filter": [{
            "field": "text",
            "relation": "contains",
            "values": ["zika"]
        }],
        "unnest": [{
            "hashtags": "tag"
        }],
        "group": {
            "by": [{
                "field": "tag"
            }],
            "aggregate": [{
                "field": "*",
                "apply": {
                    "name": "count"
                },
                "as": "count"
            }]
        },
        "select": {
            "order": ["-count"],
            "limit": 10,
            "offset": 0
        },
        "type": "nonSlicing"
    }],
    "option": {
        "sliceMillis": 200
    }
}
```

#### Proposal 2
[...]

PR #433, opened 2017-11-09
[Issue #432] Limit number of results returned

haochen07:
This PR related to Issue #432 

**Note**
- Updated json parser to parse limit number.
- Checked the number of results each time before returning them. Once reaching the limit, send `Done` message and stop querying.

**Test**
- Added test case to Json parser and cloudberry client.
- All unit test cases pass.
- Tested on current TwitterMap.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
@@ -264,7 +264,7 @@ object JSONParser {
   implicit val exeOptionReads: Reads[QueryExeOption] = (
     (__ \ QueryExeOption.TagSliceMillis).readNullable[Int].map(_.getOrElse(-1)) and
       (__ \ QueryExeOption.TagContinueSeconds).readNullable[Int].map(_.getOrElse(-1)) and
-      (__ \ QueryExeOption.TagLimit).readNullable[Int].map(_.getOrElse(-1))
+      (__ \ QueryExeOption.TagLimit).readNullable[Int].map(_.getOrElse(Int.MaxValue))
     ) (QueryExeOption.apply _)
 
 
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
@@ -24,13 +24,19 @@ class JSONParser extends IJSONParser {
     * Finally, calls [[QueryValidator]] to validate the correctness of this query.
     */
   override def parse(json: JsValue, schemaMap: Map[String, AbstractSchema]): (Seq[Query], QueryExeOption) = {
-    val option = (json \ "option").toOption.map(_.as[QueryExeOption]).getOrElse(QueryExeOption.NoSliceNoContinue)
-    val query = (json \ "batch").toOption match {
+    val (processedJson, option) = extractQueryExeOption(json)
+
+
+    println(processedJson.toString())
+    println(option.toString)
+
+
+    val query = (processedJson \ "batch").toOption match {
       case Some(groupRequest) => groupRequest.validate[Seq[UnresolvedQuery]] match {
         case js: JsSuccess[Seq[UnresolvedQuery]] => js.get
         case e: JsError => throw JsonRequestException(JsError.toJson(e).toString())
       }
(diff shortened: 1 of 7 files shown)


haochen07:
Sorry, I don't have authority to add reviewers. @JavierJia @waans11 @baiqiushi You can take a look at this ongoing PR.

JavierJia:
LGTM. I will let @waans11 and @baiqiushi to review the details.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -56,6 +56,41 @@ object JSONParser {
     resolved
   }
 
+  /**
+    * Parses json request and extract [[QueryExeOption]], then removes all option-related field from requests.
+    *
+    * @param json
+    * @return
+    */
+  def extractQueryExeOption(json: JsValue): (JsValue, QueryExeOption) = {
```
JavierJia: the logic of this function seems hard to follow. Could you elaborate the logic in the ScalaDoc? 
And if the logic is more appropriate in the `if else` block, we'd better write the logic in that way than the `case` way.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Query.scala
```diff
@@ -15,15 +15,23 @@ trait IWriteQuery extends IQuery {
 
 }
 
-case class QueryExeOption(sliceMills: Int,
-                          continueSeconds: Int)
+case class JsonRequestOption(sliceMills: Int,
```
JavierJia: I think this is still the `QueryExeOption`. It is the parsed object result, like `Query`, not the `JSON`, right?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Query.scala
```diff
@@ -15,15 +15,23 @@ trait IWriteQuery extends IQuery {
...
 
-object QueryExeOption {
-  val NoSliceNoContinue = QueryExeOption(-1, -1)
+object JsonRequestOption {
+  val NoSliceNoContinue = JsonRequestOption(-1, -1)
   val TagSliceMillis = "sliceMillis"
   val TagContinueSeconds = "continueSeconds"
 }
 
+case class QueryExeOption(sliceMills: Int,
```
JavierJia: Now I confused, why do we need this two almost identical options?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
haochen07: The parsing logic is
- if there is no `option`, treat queries by the original way (meaning if there is `limit` in the query, it will be executed in database statement not controlled by cloudberry, just as it is today)

- if there is `option`,

  - for `batch` queries, if `limit` presents, throw errors. Because now twittermap use `limit` in `batch` (the point we were missing in the previous discussion), there is one specialization for this: we enable `aggregate` and `limit` present together.

  - for single query, if `limit` presents, this is the case we are targeting. We allow two possible APIs.
```
{
    "select" : {
        "limit": 100
    }
}
```
and 
```
{
    "select" : {
        "order" : [ "-create_at"],  // required
        "limit": 100,                       // required
        "offset" : 0,                       // required
        "field": ["create_at", "id"] // optional
    }
}
```
for the first one, we remove `select` field after parsing the `limit` value to `QueryExeOption` and for the second one we take the limit value and reassign `Int.MaxValue` to `limit`. This way we keep the format constrains in `select` statement and parse the `limit` as well.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/schema/Query.scala
haochen07: Because the `limit` is in `select` field somewhere in the query, and all other options are in the `option` field. We need a case class to parse the `option` field and we also need an object that contains all options including `limit`. That is the intuitive.

haochen07:
Remove the parsing part to next PR. This one only contains the control logic. Temporarily put the `limit` API into `option`.

haochen07:
@waans11 and @baiqiushi , do you have any comments or something looks unfamiliar and need me to explain for this PR?

JavierJia:
@haochen07 Let's merge this PR and move on to the second part.

PR #441, opened 2017-11-29
[Issue #432] parse limit setting from requests

haochen07:
Related to issue #432 .

**Note**
- parse `limit` setting from JSON requests without changing current cloudberry API.
- limit control only allowed for single slicing queries.

**Testing**
Unit tests pass. Not affected current TwitterMap.


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -123,7 +123,7 @@ class BerryClient(val jsonParser: JSONParser,
 
       if (nextInterval.toDurationMillis <= 0 || hasEnoughResults(mergedResults, targetLimits)) {
         val returnedResult =
-          if (mergedResults.head.value.size > targetLimits) Seq(JsArray(mergedResults.head.value.take(targetLimits)))
+          if (targetLimits.nonEmpty && mergedResults.head.value.size > targetLimits.get) Seq(JsArray(mergedResults.head.value.take(targetLimits.get)))
```
JavierJia: `take` function can return less result if there are not enough results inside the list. So another easier way of writing the code is 
```
val limitResultOpt =targetLimits.map(limit => Seq(JsArray(mergedResults.head.value.take(limit))))
val returnedResult = limitResultOpt.getOrElse(mergedResult)
```

which is cleaner than the `if else` clause.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/BerryClient.scala
```diff
@@ -39,7 +39,7 @@ class BerryClient(val jsonParser: JSONParser,
 
   private case class QueryGroup(ts: DateTime, curSender: ActorRef, queries: Seq[QueryInfo], postTransform: IPostTransform)
 
-  private case class Initial(ts: DateTime, sender: ActorRef, targetMillis: Long, targetLimits: Int, queries: Seq[Query], infos: Map[String, DataSetInfo], postTransform: IPostTransform)
+  private case class Initial(ts: DateTime, sender: ActorRef, targetMillis: Long, targetLimits: Option[Int], queries: Seq[Query], infos: Map[String, DataSetInfo], postTransform: IPostTransform)
```
JavierJia: can we add another sanity check about the logic that no query groups are allowed when limit appears when we receive the query?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -56,6 +57,27 @@ object JSONParser {
...
+        val limit = (json \ "select" \ "limit").as[Int]
+        val updatedSelectJson = (json \ "select").as[JsObject] ++ Json.obj("limit" -> Int.MaxValue)
+        val updatedJson = json.as[JsObject] ++ Json.obj("select" -> updatedSelectJson)
+        (updatedJson, QueryExeOption(option.sliceMills, option.continueSeconds, Some(limit)))
+      }
+    } else {  // batch slicing query
+      if ((json \\ "limit").isEmpty || ((json \\ "limit").nonEmpty && (json \\ "aggregate").nonEmpty)) {
+        (json, QueryExeOption(option.sliceMills, option.continueSeconds, None))
+      } else {
+        throw JsonRequestException("Batch Requests cannot contain \"limit\" field")
```
JavierJia: we should let the user know the request is wrong, right? 
but I think it requires a more framework change. let's leave a TODO here.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -56,6 +57,27 @@ object JSONParser {
     resolved
   }
 
+  def parseLimit(json: JsValue, option: QueryExeOption): (JsValue, QueryExeOption) = {
+    if (option.sliceMills <= 0) { // non-slicing query
+      (json, QueryExeOption(option.sliceMills, option.continueSeconds, None))
```
JavierJia: you can use `return` to reduce the nested else to reduce the complexity.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -56,6 +57,27 @@ object JSONParser {
     resolved
   }
 
+  def parseLimit(json: JsValue, option: QueryExeOption): (JsValue, QueryExeOption) = {
```
JavierJia: this function contains too many nested if else. I would suggest rewriting the logic by adding `return`s to at least eliminate the nested branches.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -56,6 +57,27 @@ object JSONParser {
...
+    if (option.sliceMills <= 0) { // non-slicing query
+      (json, QueryExeOption(option.sliceMills, option.continueSeconds, None))
+    } else if (option.sliceMills > 0 && (json \ "batch").toOption.isEmpty) { // single slicing query
+      if ((json \\ "limit").isEmpty) {
+        (json, QueryExeOption(option.sliceMills, option.continueSeconds, None))
+      } else {
+        val limit = (json \ "select" \ "limit").as[Int]
+        val updatedSelectJson = (json \ "select").as[JsObject] ++ Json.obj("limit" -> Int.MaxValue)
```
JavierJia: if we only `++`, the original `limit` property is still there, right?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
```diff
@@ -263,7 +285,8 @@ object JSONParser {
   implicit val exeOptionReads: Reads[QueryExeOption] = (
     (__ \ QueryExeOption.TagSliceMillis).readNullable[Int].map(_.getOrElse(-1)) and
       (__ \ QueryExeOption.TagContinueSeconds).readNullable[Int].map(_.getOrElse(-1)) and
-      (__ \ QueryExeOption.TagLimit).readNullable[Int].map(_.getOrElse(Int.MaxValue))
+      // this "limit" value will be discarded, real limit value is parsed from request
```
JavierJia: so we should remove it?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
haochen07: No, it will be overwritten. I had a unit test case in JsonParserTest to test it.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
haochen07: Unless we have a case class that contains only `sliceMillis` and `ContinueSeconds`. This is why last time I create another whatever case class to parse it.

JavierJia: Got it.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/JSONParser.scala
JavierJia: Got it.

haochen07: in "parse limit in single query"

JavierJia:
LGTM 👍

JavierJia:
@waans11 @baiqiushi do you have any questions about this PR?

waans11:
No, unlike the previous Web socket change, this one is more Scala based, so it's hard to understand. :-)

JavierJia:
so, should we merge this one?

PR #447, opened 2017-12-06
Changed two map icons

waans11:
- Changed Countmap and Pointmap icons.
twittermap 2017-12-06 10-07-01


chenlica:
It will be good to add a screenshot after the change (for our records).

PR #448, opened 2017-12-06
Fixing map-switching bug

DeyuChen:
Fixed polygon displaying issue when switching between pointmap and countmap


examples/twittermap/web/public/javascripts/map/controllers.js
```diff
@@ -430,6 +435,22 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common','clou
...
+      }
+      if ($scope.polygons.cityPolygons) {
+        $scope.map.removeLayer($scope.polygons.cityPolygons);
+      }
+      if ($scope.polygons.stateUpperPolygons) {
+        $scope.map.removeLayer($scope.polygons.stateUpperPolygons);
+      }
+      if ($scope.polygons.countyUpperPolygons) {
+        $scope.map.removeLayer($scope.polygons.countyUpperPolygons);
+      }
```
waans11: These lines that remove layers are duplicated in the cleanCountMap() function. From cleanCountMap() function, can you extract these common layer removal parts and create another function and use it in the both places?

examples/twittermap/web/public/javascripts/map/controllers.js
```diff
@@ -642,6 +672,14 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common','clou
...
           });
           setCenterAndBoundry($scope.geojsonData.county.features);
+          if ($scope.status.logicLevel === 'county'){
+            $scope.map.addLayer($scope.polygons.stateUpperPolygons);
+            $scope.map.addLayer($scope.polygons.countyPolygons);
+            if (cloudberry.parameters.maptype === "countmap"){
+              cloudberry.parameters.geoLevel = $scope.status.logicLevel;
+              cloudberry.query(cloudberry.parameters);
+            }
+          }
```
waans11: We don't need to consider "statePolygons" or "cityPolygons" at this time?

examples/twittermap/web/public/javascripts/map/controllers.js
baiqiushi: For these remove layers actions, I think the cleanCountMap() function already did. And every time the setInfoControlPointMap() is called, the cleanCountMap() must have been called previously. So I guess these lines are duplicate.

DeyuChen: These layers should be removed when we switch from pointmap to countmap, too.

examples/twittermap/web/public/javascripts/map/controllers.js
DeyuChen: This is called when county.json file is successfully loaded. "statePolygons" layer should be added when state.json is loaded and $scope.status.logicLevel === 'state'. If the logic level is city, both state and county layers will not be added. Instead, I call loadCityJsonByBound(onEachFeature) to add city polygon layer.

examples/twittermap/web/public/javascripts/map/controllers.js
waans11: Teyu, why that's the case?

examples/twittermap/web/public/javascripts/map/controllers.js
waans11: Got it.

examples/twittermap/web/public/javascripts/map/controllers.js
DeyuChen: Because right now pointmap still add those polygons to the map, but with lighter color.

DeyuChen: I added a new function removePolygonLayers() that will be used by both cleanCountMap() and cleanPointMap().

PR #449, opened 2017-12-06
Change three icons

waans11:
- Changed three icons (made by Aaron Zhang)
twittermap 2017-12-06 10-56-45


chenlica:
In our production system we only show two icons (PointMap and CountMap), right?  The HeatMap icon will be added in the future when the feature is ready.

waans11:
I am going to disable it temporarily in the system.

PR #450, opened 2017-12-07
Make maxTextMessageSize of websocket server configurable

haochen07:
Quick fix on configuring the `maxTextMessageSize` setting instead of hard coding.


examples/twittermap/web/app/websocket/WebSocketFactory.scala
```diff
@@ -5,10 +5,10 @@ import org.eclipse.jetty.websocket.client.WebSocketClient
 
 class WebSocketFactory {
 
-  def newClient(): WebSocketClient = {
+  def newClient(maxTextMessageSize: Int): WebSocketClient = {
     val socket = new WebSocketClient()
-    //TODO Should be configurable in later PR.
-    socket.setMaxTextMessageBufferSize(5*1024*1024)
+    socket.setMaxTextMessageBufferSize(maxTextMessageSize)
+    socket.getPolicy.setMaxTextMessageSize(maxTextMessageSize)
```
waans11: What's the difference between these two lines? Just curious.

examples/twittermap/web/conf/application.conf
```diff
@@ -80,6 +80,7 @@ cloudberry.register = "http://localhost:9000/admin/register"
 #cloudberry.ws = "ws://textdb.ics.uci.edu:9000/ws"
 cloudberry.ws = "ws://localhost:9000/ws"
 
+maxTextMessageSize = 5242880 # 5MB
```
waans11: Could you put a comment above like the following?
Maximum allowable text message size of the web socket

examples/twittermap/web/conf/production.conf
```diff
@@ -79,5 +79,6 @@ bounded-mailbox {
 
 cloudberry.register = "http://localhost:9000/admin/register"
 cloudberry.ws = "ws://localhost:9000/ws"
+maxTextMessageSize = 5242880 # 5MB
```
waans11: Comment here, too.

haochen07:
I tested this change on the new PointMap locally. I am not sure if this is the test case we had used when we noticed the issue of message size. @baiqiushi can you check it ?

examples/twittermap/web/app/websocket/WebSocketFactory.scala
haochen07: http://www.eclipse.org/jetty/javadoc/9.4.4.v20170414/org/eclipse/jetty/websocket/api/WebSocketPolicy.html#setMaxTextMessageBufferSize-int-

Seems the `TextMessageSize` is what we want to configure. But I don't have large data set to test so I keep the two changes we had previously.

haochen07:
This PR is a quick fix for message size configuration, the issue @baiqiushi had mentioned. You could have a review on it.

chenlica:
@baiqiushi and @JavierJia : please review and finish this PR.

PR #453, opened 2017-12-08
A debug version on testing websocket overhead

haochen07:
This is the **debug** version to test websocket server overhead when we switch 3-layer to 4-layer architecture. We could keep it in a debug branch and test it in production when available.

I have tested some of the keywords search locally and below is the experimental data.

screen shot 2017-12-07 at 8 29 01 pm

### DIFF
--- a/examples/twittermap/web/public/javascripts/common/services.js
+++ b/examples/twittermap/web/public/javascripts/common/services.js
@@ -355,9 +355,6 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
 
               ws.send(sampleJson);
               ws.send(batchWithPartialGeoRequest);
-
-              console.log(new Date().getTime() + "\t" + parameters.keywords);
-
             }
             break;
             
@@ -416,6 +413,8 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
             }));
 
             ws.send(pointsJson);
+            console.time("sample response time");
+
             ws.send(pointsTimeJson);
             break;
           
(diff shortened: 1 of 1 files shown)


JavierJia:
I would suggest use the `console.time` (in [stackoverflow](https://stackoverflow.com/a/1975103/2598198)) to better record the round-trip time. 

Since we've already had the point map, could we measure its time?

haochen07:
We can measure point map but since we don't have point map for 3-layer architecture, we cannot draw any conclusion.

haochen07:
I removed all intermediate result times and only kept the final one. Note that for the keyword "This is us", 4-layer version returned more results thus its response time exceeded a lot. Generally, the 4-layer one present some overhead.

screen shot 2017-12-08 at 10 24 41 am

JavierJia:
The sample request (the one pull the id of the tweets to render them in the sidebar) is of the same type with the pointmap.  You could try that one and possibly increase the limit number and date.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -413,6 +413,8 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
             }));
 
             ws.send(pointsJson);
+            console.time("sample response time");
```
haochen07: This is the place we send sample Json in point map, right?

JavierJia: it should be. @baiqiushi can you confirm?

haochen07:
The new commit made the switch of 3-layer and 4-layer architecture configurable.
I used point map with Drum for both test environment. The only difference is the architecture layer.
A quick experiment result is shown below.


screen shot 2017-12-09 at 11 42 22 am

haochen07:
I repeated the same experiment 5 times. The new results below showed some consistency. It seems that the overhead is not significant in general.

screen shot 2017-12-11 at 9 53 27 am

chenlica:
The results are very good!  Seems the overhead of the 4-tier architecture is not much.  I feel comfortable with the "3->4 switch."

haochen07:
This PR aims at testing the overhead of 4-layer architecture. Since we have drawn the conclusion, this PR is no longer useful.

Issue #464, opened 2017-12-20
Add "is Null" and "is not Null" predicates to the filtering API


baiqiushi:
Currently, Cloudberry query language model does not support "is Null" or "is not Null" filtering predicates. But AsterixDB and most Database systems support these selection semantics, so, it's natural to add them.

In addition to current [Cloudberry Query Model](https://docs.google.com/document/d/1-xTBxdIg_y_nHMpztZdEG3i_IOpeKhwIT6UQJVDL0UM/edit?usp=sharing), add 2 `filter` predicates.

- `IS NULL`
  * **filter grammer**: `isNull`
  * **Applied DataType**: `Boolean`, `Number`, `Point`, `Time`, `String`, `Text`
  * **Implementation**:
      -- SQL++ = `IS UNKNOWN`
      -- MySQL = `IS NULL`
- `IS NOT NULL`
  * **filter grammer**: `isNotNull`
  * **Applied DataType**: `Boolean`, `Number`, `Point`, `Time`, `String`, `Text`
  * **Implementation**:
      -- SQL++ = `IS NOT UNKNOWN`
      -- MySQL = `IS NOT NULL`

**Example**:
```
"filter": [
   {
     "field": "fieldP",
     "apply": "func",
     "relation": "isNull|isNotNull",
     "value": []
   }
]
```

PR #465, opened 2017-12-31
Add isNull and isNotNull filter relations

baiqiushi:
Solve issue #464 
In addition to current [Cloudberry Query Model](https://docs.google.com/document/d/1-xTBxdIg_y_nHMpztZdEG3i_IOpeKhwIT6UQJVDL0UM/edit?usp=sharing), add 2 `filter` predicates.

- `IS NULL`
  * **filter grammer**: `isNull`
  * **Applied DataType**: `Boolean`, `Number`, `Point`, `Time`, `String`, `Text`
  * **Implementation**:
      -- SQL++ = `IS UNKNOWN`
      -- MySQL = `IS NULL`
- `IS NOT NULL`
  * **filter grammer**: `isNotNull`
  * **Applied DataType**: `Boolean`, `Number`, `Point`, `Time`, `String`, `Text`
  * **Implementation**:
      -- SQL++ = `IS NOT UNKNOWN`
      -- MySQL = `IS NOT NULL`

**Example**:
```
"filter": [
   {
     "field": "fieldP",
     "apply": "func",
     "relation": "isNull|isNotNull",
     "value": []
   }
]
```
**NOTE: `No implementation for another 2 QueryGenerators: PostgreSQL & SparkSQL`**

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
@@ -155,44 +155,31 @@ abstract class AsterixQueryGenerator extends IQLGenerator {
 
   //TODO possibly using /*+ skip-index */ hint if the relation selectivity is not high enough
   protected def parseFilterRelation(filter: FilterStatement, fieldExpr: String): String = {
-    filter.relation match {
-      case Relation.isNull => {
-        filter.field.dataType match {
-          case DataType.Bag => ???
-          case DataType.Hierarchy =>
-            throw new QueryParsingException("the Hierarchy type doesn't support any relations.")
-          case _ =>
-            s"$fieldExpr is unknown"
-        }
-      }
-      case Relation.isNotNull => {
-        filter.field.dataType match {
-          case DataType.Bag => ???
-          case DataType.Hierarchy =>
-            throw new QueryParsingException("the Hierarchy type doesn't support any relations.")
-          case _ =>
-            s"$fieldExpr is not unknown"
-        }
-      }
-      case _ => {
-        filter.field.dataType match {
-          case DataType.Number =>
-            parseNumberRelation(filter, fieldExpr)
-          case DataType.Time =>
-            parseTimeRelation(filter, fieldExpr)
-          case DataType.Point =>
-            parsePointRelation(filter, fieldExpr)
-          case DataType.Boolean => ???
-          case DataType.String => parseStringRelation(filter, fieldExpr)
-          case DataType.Text =>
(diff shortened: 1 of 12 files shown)


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
```diff
@@ -183,6 +183,12 @@ abstract class AsterixQueryGenerator extends IQLGenerator {
       case Relation.inRange => {
         s"$fieldExpr >= ${typeImpl.datetime}('${filter.values(0)}') and $fieldExpr < ${typeImpl.datetime}('${filter.values(1)}')"
       }
+      case Relation.isNull => {
+        s"$fieldExpr is unknown"
+      }
+      case Relation.isNotNull => {
+        s"$fieldExpr is not unknown"
+      }
```
JavierJia: Is it better to move the repeated code from line 186 to 191 in different functions to a unified place? E.g., could we handle the `isNull` or `isNotNull` in the entry caller `parseFilterRelation`?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryValidator.scala
```diff
@@ -169,6 +181,9 @@ object QueryValidator {
   }
 
   def validateTextRelation(relation: Relation, values: Seq[Any]): Unit = {
+    if (relation == Relation.isNull || relation == Relation.isNotNull) {
+      return
+    }
```
JavierJia: Similar to the previous comments, if the `null` logic is handled in a unified place, we should not need to validate it here.

chenlica:
@JavierJia Thanks for giving a detailed review for the first scala PR of @baiqiushi :-)

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/AsterixQueryGenerator.scala
baiqiushi: @JavierJia Sorry, what do you mean by `JSONRequest` changes? Do you mean `JSONParser`?
For the unification part, I agree, I just didn't know how to do it without changing too much code, I will try to think of one way. Thank you.

chenlica:
I saw some changes that are simply related to reformatting.  It will be good to separate such changes from logical changes.  We can do them as a separate PR.  Just some thoughts ...

baiqiushi:
Yeah, there are only 2 empty lines which are modified because that function was in my earlier commit, but after taking Jianfeng's advice to unify the logic to a higher place, that function was reverted to original version, those two empty lines were only reverted because it's more coherent to the format of other original code, so I left them there.

Issue #443, opened 2017-12-04
Remove RequestRouter Actor


haochen07:
After Drum refactoring #442 , the `Berryclient` can separate non-slicing and slicing queries to different resolvers, which makes the `RequestRouter` actor redundant. We could remove it to reduce one hop of actor flow.

PR #466, opened 2018-01-05
Simplify the RequestRouter Actor

JavierJia:
The Drum refactoring #442 makes the `BerryClient` can handle both non-slicing and slicing queries in a single actor. We don't need the `RequestRouter` actor to create separate actors. However, we still keep this router to log client information.

Issue: #443


haochen07:
LGTM. Maybe you need to change the unit test case for `RequestRouter` as well. Now the building process fails.

JavierJia:
It is fixed now 😄

Issue #467, opened 2018-01-05
Add "id" and "category" parameter to replace current "key" parameter of Query API.


baiqiushi:
**Problem:**
    Previously, Cloudberry Query API was designed with a keyword called "transform", and inside it there is one function called "wrap" implemented. The purpose of "wrap" it to wrap the result into a package and identify it with a "key". Therefore, "key" has been used by the Frontend user (i.e. TwitterMap) as an request ID to identify the result belongs to which query it sent. However, as issue #463 indicates, now we need not only a request ID to identify a query, but also a query "category" as an unique name for our backend server to create proper number of actors solving queries.

**Solution:**
    Here comes the solution, we use 2 parameters inside the "wrap" function of keyword "transform", namely "id" and "category". The "id" serves as an identification of each query and its corresponding result. The "category" serves as a classification of each query for the Cloudberry to schedule its resources more properly.

Issue #463, opened 2017-12-16
Create a downstream actor for each "category" query.


baiqiushi:
**Problem:** 
    Currently, `berryClient` only creates two downstream actors, one actor for non-slicing queries and the other actor (named `stream`) for slicing queries.  The problem is it will not allow any frontend user to send two different `category`s of slicing queries simultaneously, because the single `stream` actor will drop any previous slicing queries once it receives a new slicing query.

**Solution:**
    Before this solution, issue #467 should be solved first, because this issue needs the `category` parameter as a prerequisite.
    Change the logic of creating slicing-query actor to creating one actor which will be named with the value of `category` parameter inside the query's `transform` keyword, therefore, for each different `category` query, there will be one actor solving it.

chenlica:
Clarification: here a "key" means a "type" of query.

@baiqiushi Please also include the original issue number and do cross referencing between these two issues.

baiqiushi:
@chenlica  Sorry, I don't know there's any related issue exists there.

PR #470, opened 2018-01-07
Add id and type to replace current key of Query

baiqiushi:
**Problem #467 :**
    Previously, Cloudberry Query API was designed with a keyword called "transform", and inside it there is one function called "wrap" implemented. The purpose of "wrap" it to wrap the result into a package and identify it with a "key". Therefore, "key" has been used by the Frontend user (i.e. TwitterMap) as an request ID to identify the result belongs to which query it sent. However, as issue #463 indicates, now we need not only a request ID to identify a query, but also a query "type" as an unique name for our backend server to create proper number of actors solving queries.

**Solution:**
    Here comes the solution, we use 2 parameters inside the "wrap" function of keyword "transform", namely "id" and "type". The "id" serves as an identification of each query and its corresponding result. The "type" serves as a classification of each query for the Cloudberry to schedule its resources more properly.

**Problem #463 :** 
    Currently, `berryClient` only creates two downstream actors, one actor for non-slicing queries and the other actor (named `stream`) for slicing queries.  The problem is it will not allow any frontend user to send two different `type`s of slicing queries simultaneously, because the single `stream` actor will drop any previous slicing queries once it receives a new slicing query.

**Solution:**
    Before this solution, issue #467 should be solved first, because this issue needs the `type` parameter as a prerequisite.
    Change the logic of creating slicing-query actor to creating one actor which will be named with the value of `type` parameter inside the query's `transform` keyword, therefore, for each different `type` query, there will be one actor solving it.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IPostTransform.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IPostTransform.scala
@@ -4,10 +4,12 @@ import play.api.libs.json.JsValue
 
 
 trait IPostTransform {
+  def category: String
   def transform(jsValue: JsValue): JsValue
 }
 
 case object NoTransform extends IPostTransform {
+  override def category: String = "default"
   override def transform(jsValue: JsValue): JsValue = jsValue
 }
 
--- a/cloudberry/neo/app/actor/RequestRouter.scala
+++ b/cloudberry/neo/app/actor/RequestRouter.scala
@@ -36,7 +36,7 @@ class RequestRouter (berryClientProp: Props, config: Config, requestHeader: Requ
     (requestBody \ "transform").asOpt[JsValue] match {
       case Some(t) =>
         (t \ "wrap").asOpt[JsValue] match {
-          case Some(w) => WrapTransform((w \"key").as[String])
+          case Some(w) => WrapTransform((w \"id").as[String], (w \"category").as[String])
           case None => NoTransform
         }
       case None => NoTransform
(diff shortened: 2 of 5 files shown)


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IPostTransform.scala
```diff
@@ -4,10 +4,12 @@ import play.api.libs.json.JsValue
 
 
 trait IPostTransform {
+  def category: String
```
JavierJia: why do we need a `category` property in the interface?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IPostTransform.scala
```diff
@@ -4,10 +4,12 @@ import play.api.libs.json.JsValue
...
   def transform(jsValue: JsValue): JsValue
 }
 
 case object NoTransform extends IPostTransform {
+  override def category: String = "default"
```
JavierJia: If you need some class have the `category`, we should add a `trait` or `class` that inherits from the `IPostTransform` to be explicit.  E.g.
```
case class CategoricalWrap extends IPostTransform {
    def category: String
}
```

baiqiushi: Yes, we can achieve that in this way, but if so, the `handleRequest` of `BerryClient` parameter type has to be changed to `CategoricalWrap` as well. I am wondering whether it is good to change use cases of an interface instead of add a feature inside the interface.
What do you think? @JavierJia

chenlica:
Is it better to split this PR into two, or it must be "2 in 1"?

JavierJia:
@baiqiushi, there are some tests failing, please check the *Detiails* of the *Travis* for the details.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/datastore/IPostTransform.scala
JavierJia: Interfaces should be succinct and self-explanable. We are refactoring towards a "correct" direction, not an "easier" direction. Most of the time, the "correct" will be "easier" ultimately.

PR #471, opened 2018-01-08
Add "id" and "category" parameter to replace current "key" parameter of Query API

baiqiushi:
Issue #467 
**Problem:**
    Previously, Cloudberry Query API was designed with a keyword called "transform", and inside it there is one function called "wrap" implemented. The purpose of "wrap" it to wrap the result into a package and identify it with a "key". Therefore, "key" has been used by the Frontend user (i.e. TwitterMap) as an request ID to identify the result belongs to which query it sent. However, as issue #463 indicates, now we need not only a request ID to identify a query, but also a query "category" as an unique name for our backend server to create proper number of actors solving queries.

**Solution:**
    Here comes the solution, we use 2 parameters inside the "wrap" function of keyword "transform", namely "id" and "category". The "id" serves as an identification of each query and its corresponding result. The "category" serves as a classification of each query for the Cloudberry to schedule its resources more properly.

### DIFF
--- a/cloudberry/neo/app/actor/RequestRouter.scala
+++ b/cloudberry/neo/app/actor/RequestRouter.scala
@@ -4,7 +4,7 @@ import akka.actor.{Actor, ActorLogging, Props}
 import akka.stream.Materializer
 import edu.uci.ics.cloudberry.zion.actor.BerryClient._
 import edu.uci.ics.cloudberry.zion.common.Config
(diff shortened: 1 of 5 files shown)


baiqiushi:
@chenlica, I split it into 2 prs, and this is only for issue #467 .
@JavierJia, what about this time?

Issue #468, opened 2018-01-05
Enable query slicing for PointMap


baiqiushi:
**Problem:**
Currently, `PointMap` are using 2 single queries: 1) query points; 2) query "Time Series Bar";
Why it cannot use a `batch` query? Because `PointMap` still needs a `limit` feature to constrain the size of result returns in case not to crash the browser.
Why it cannot enable the 2 single queries both with `query-slicing`? Because without issue #463 solved, Cloudberry will abandon one of the 2 queries which arrives earlier.

**Solution:**
Therefore, after solving issue #463 , it will be easy to enable the 2 single queries both with `query-slicing`.

baiqiushi:
#473

PR #473, opened 2018-01-11
Enable query slicing for PointMap

baiqiushi:
#468 

giphy


chenlica:
@waans11 : can you verify this slicing feature on PointMap?  If so, let's merge it and deploy it to the backup server!

PR #476, opened 2018-01-18
Make parameters of Pointmap configurable

baiqiushi:
Before enabling Pointmap in our production system, @chenlica suggests that we tune the two parameters: sampling time range which is `pointmapSamplingDayRange` and returned result records limit which is `pointmapSamplingLimit` both on Backup server and Production server to make the user experience more reasonable.

This PR is to make these 2 parameters configurable in the `application.conf ` file of Twittermap.

### DIFF
--- a/examples/twittermap/web/conf/application.conf
+++ b/examples/twittermap/web/conf/application.conf
@@ -116,5 +116,7 @@ us.city.path = "/public/data/city.sample.json"
 tweet.url = "http://twitterMapIndex-search-proxy.herokuapp.com/search/tweets?q=%s"
 
 # Pointmap parameters
-pointmap.samplingDayRange = 30
-pointmap.samplingLimit = 5000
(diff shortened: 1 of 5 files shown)


examples/twittermap/web/conf/application.conf
```diff
@@ -115,3 +115,6 @@ us.city.path = "/public/data/city.sample.json"
 
 tweet.url = "http://twitterMapIndex-search-proxy.herokuapp.com/search/tweets?q=%s"
 
+# Pointmap parameters
```
chenlica: Add comments to explain the purpose of these two parameters.

waans11: In addition, you need to explain what these parameters do. Also, if the default value is used in case no parameters are provided, we can comment the following two parameters (e.g., querySliceMills and cacheThreshold).

PR #477, opened 2018-01-22
Use '/' rather than '%' to check HTTP status code

guori12321:


JavierJia:
You can click the Travis `Details` button to see why the build is failed. 

It seems the `RequestRouterTest` test has failed. 
```
[error]   x forward request
[error]    'null' is not equal to '{"sender":"$a"}' (RequestRouterTest.scala:109)
```
It may because the original test was relying on that incorrect code. @guori12321 can you give it a try to fix the test? 😄

guori12321:
Sure. Let me check...

waans11:
@JavierJia @baiqiushi Please check this patch set again. Let's merge if nothing happens for the build.

chenlica:
@guori12321 @ruig2 @baiqiushi status of this PR?  reject it or finish it?

ruig2:
@baiqiushi I tested on my local env and this PR works fine.

In fact, the returned value is **never used**. The only call site of this `up` method is at https://github.com/ISG-ICS/cloudberry/blob/master/examples/twittermap/web/app/controllers/TwitterMapApplication.scala#L74-L76  , where the returned value of `Await.result(register, 1 minutes)` is not used at all.

**More details**
More details about the `up` method (https://github.com/ISG-ICS/cloudberry/pull/477/files#diff-87695521c7c23c62db8dc37914bfb7e0L12): the HTTP response of it would always be `400`, so both `response.status / 100 == 2` and `response.status % 100 == 2` are both false. The 400 means the following 4 requests are bad:

```
[info] application - Bad Request:Register Denied: dataset twitter.ds_tweet already existed.
[info] application - Bad Request:Register Denied: dataset twitter.dsCityPopulation already existed.
[info] application - Bad Request:Register Denied: dataset twitter.dsCountyPopulation already existed.
[info] application - Bad Request:Register Denied: dataset twitter.dsStatePopulation already existed.
```

And they are bad because that we now initialize the database via the script `ingestTwitterToLocalCluster.sh` (which is triggered by `ingestAllTwitterToLocalCluster.sh`). You can see that the 4 above datasets are all created in the script when we set up the dev environment. So an `already existed` error will be caught.

All in all, the returned value is never used because the `already existed` error should be ignored here.

PR #481, opened 2018-01-27
Apply the current AsterixDB's API change to TwitterMap

waans11:
- Apply the current AsterixDB's API change (the current release of AsterixDB uses SQLPP by default).


baiqiushi:
From @waans11 feedback quotation:
For that PR, $host was originally a path (e.g., http://localhost:19002/aql). But we used $host to indicate an IP. And I think 127.0.0.1 is OK since each file is ingesting a sample data to the localhost.

Issue #480, opened 2018-01-27
Wrap conf parameters of Twittermap application into objects


baiqiushi:
**Problem**
Twittermap application uses `application.conf` file to store configuration parameters. When the `TwitterMapApplication.scala` starts up, it loads the parameters into web html files and javascript objects. Currently, these parameters are transferred in a way all aligned together between files and codes. Apparently, it's not a good practice.

**Solution**
Whether it's possible is not quite clear yet. When `TwitterMapApplication.scala` initialize the `index.scala.html` and `main.scala.html` files, wrap the parameters into several objects based on their purpose and usage.

**Related technology**
 - Scala
 - Play Framework

chenlica:
@ChankiWu Good luck to your first backend task!

PR #483, opened 2018-02-07
Wrap conf parameters of Twittermap application into objects-to fix the issue #480

ChankiWu:
Has finished the task and tested running the project !


chenlica:
Given the fact this is the first backend PR of @ChankiWu, I would like to get involved to do a careful review.  

A few high-level comments:

- The subject should cite the issue number.  See other PRs;
- In the description, explain the main idea of the changes;
- Get rid of the unnecessary format-related changes since they made the PR harder to review.

@baiqiushi can do the review first.  After that, we do a review together.

PR #485, opened 2018-02-10
[Issue #480] Wrap conf parameters of Twittermap application into object

ChankiWu:
Optimize the structure of config parameters to pass among files. Wrap config parameters of Twittermap application into object and use this object to get corresponding values.
#480


chenlica:
LGTM.

Issue #493, opened 2018-03-07
Change pinmap's method of getting results from cloudberry directly instead of Twitter API.


ChankiWu:
The previous way of getting results to show in Front-end of pinmap is sending query to Twitter API. Now changing this into sending query directly to middleware cloudberry.

chenlica:
I don't want to be a blocker.  Please let @baiqiushi and @waans11 do the reviews.

PR #494, opened 2018-03-07
[Issue #493] Pinmap gets tweet's results from cloudberry.

ChankiWu:
The previous way of getting results to show in Front-end of pinmap is sending query to Twitter API. Now changing this into sending query directly to middleware cloudberry. The whole logic basically works fine. In 
addition, there are still some issues such as no result stored in DB for few points, showing more information such as profile_img in the Front-end after getting large data sets. May need some refinements afterwards.


ChankiWu:
@chenlica Sure. Will do ! The feature is basically the same as the previous one.

ChankiWu:
Screenshots between previous one and current one.
image

image

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -266,7 +331,21 @@ angular.module('cloudberry.map')
               $scope.popUpTweet = L.popup({maxWidth:300, minWidth:300, maxHight:300});
               $scope.popUpTweet.setContent(tweetContent);
               $scope.currentMarker.bindPopup($scope.popUpTweet).openPopup();
+            });*/
```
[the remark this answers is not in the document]
ChankiWu: Yep. Thank you for pointing it out! It is because this version is not the final version of this task, so I just leave it there for the further use. The comment code is previously qiushi's code, and i am just not sure whether to delete it.

ChankiWu:
Ready to be reviewed.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -11,6 +11,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       sentimentUpperBound: 4,
       cacheThreshold: parseInt(config.cacheThreshold),
       querySliceMills: parseInt(config.querySliceMills),
+      tweetDBresult: null,
```
waans11: Let's change the name: pinMapOneTweetResult

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -41,7 +42,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     };
   })
-  .service('cloudberry', function($timeout, cloudberryConfig, MapResultCache) {
+  .service('cloudberry', function($timeout, $rootScope, cloudberryConfig, MapResultCache) {
```
waans11: Why does $rootScope need to be involved? I saw the logic (line 265-270). But, no map types uses rootScope to get a result? Can we follow that logic?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,39 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
             // unrecognized map type
             break;
         }
+      },
+
+      querypin: function(pinid){
```
waans11: The name should be precise. 

pinMapOneTweetQuery: function(tweetid) ... 

Please adjust the related parts, too.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,39 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+              filter: setpinfilter,
+              select:{
+                  order: ["-create_at"],
+                  limit: defaultSamplingSize,
+                  offset: 0,
+                  field: ["text","user.id","create_at","user.name","user.profile_image_url"]
+              },
+              transform:{
+                  wrap:{
+                      id:"pinDBtweet",
```
waans11: This ID too...

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,39 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                      category:"pinDBtweet"
+                  }
+              }
+          }));
+
+
+          //changes the id's type from string to int64.
+          pinDBresult = pinDBresult.replace(/"(\d+)"/,'[$1]');
```
waans11: What does this do? I mean, why is this needed?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -485,6 +519,10 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
           case "totalCount":
             cloudberryService.commonTotalCount = result.value[0][0].count;
             break;
+          case "pinDBtweet":
```
waans11: Name change

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -485,6 +519,10 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+            cloudberryConfig.tweetDBresult = result.value[0][0];
+            $rootScope.$emit("TweetData",cloudberryConfig.tweetDBresult);
```
waans11: This one... Why we need to use $rootScope?

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -253,19 +256,18 @@ angular.module('cloudberry.map')
...
-              $scope.popUpTweet = L.popup({maxWidth:300, minWidth:300, maxHight:300});
-              $scope.popUpTweet.setContent(tweetContent);
-              $scope.currentMarker.bindPopup($scope.popUpTweet).openPopup();
+
+            //send the query to cloudberry.
+            var passID = "" + $scope.pointIDs[i];
+            cloudberry.querypin(passID);
+
+            //receives the result.
+            $rootScope.$on("TweetData",function (event, args){
```
waans11: Maybe we can find another way?

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: ok, will change. Thanks !

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: As the comment said, the id is string and will be truncated and lost some digits. That's a critical bug I encountered when I do this task. Then we just remove the "" to let cloudberry accept. In Twitter's Database, it has two id's field and type is int and string to avoid this problem.

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: This way of communicating between pointmap.js and service.js is tried in my first PR by Qiushi, teyu and me. So I talk with Qiushi and decide to use this method to finish the communication between the two javascript files.

examples/twittermap/web/public/javascripts/map/pointmap.js
ChankiWu: So is there any suggestions ?

ChankiWu:
OK, then when would you be available next week ? Due to the Final, I will be available after next Tuesday. @waans11 If you are busy, maybe we could meet when next quarter begins.

chenlica:
It's a good idea to do it sooner than later as your memories are still
fresh.

waans11:
We may not need to meet. 

Can you check the following parts and adjust these methods to your pull request? The basic idea here is that keep tracking of the pointmap or the countmap related variables and conducts some actions.

pointmap.js: line 329 -
$scope.$watch(

countmap.js: line 444 -
$scope.$watchCollection(

ChankiWu:
@waans11 Updated as you suggest ! Thank you so much for your suggestion.

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,69 @@ angular.module('cloudberry.map')
...
-          try {
-            userLink = tweetJSON.author_url;
-          }
-          catch (e) {
-            console.log("author_url missing in this Tweet.:" + e.message);
-          }
+            var userPhotoUrl = "";
+            try {
+                userPhotoUrl= tweetJSON['user.profile_image_url'];
+                console.log("userPhotoUrl = " + userPhotoUrl);
```
baiqiushi: Please get rid of this log, only leaves log for exception cases

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,69 @@ angular.module('cloudberry.map')
...
+            }
+            catch (e){
+                console.log("user.profile_image_url missing in this Tweet.:");
+            }
 
-          var tweetLink = "";
```
baiqiushi: We can keep the tweetLink by construct the link using this pattern : https://twitter.com/{username}/status/{tweetid}

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,69 @@ angular.module('cloudberry.map')
...
-            tweetText = tweetHtml.getElementsByTagName('p')[0].innerHTML;
-          }
-          catch (e){
-            console.log("html missing in this Tweet.:" + e.message);
-          }
+            var tweetTime = "";
+            try {
+                var created_at = new Date(tweetJSON['create_at']);
+                tweetTime = created_at.toTimeString() + "\\t" + created_at.toDateString();
+                console.log("tweetTime = " + tweetTime);
```
baiqiushi: The same, get rid of the log here.

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,69 @@ angular.module('cloudberry.map')
...
-            var tweetHtml = new DOMParser().parseFromString(tweetJSON.html, 'text/html');
-            tweetText = tweetHtml.getElementsByTagName('p')[0].innerHTML;
-          }
-          catch (e){
-            console.log("html missing in this Tweet.:" + e.message);
-          }
+            var tweetTime = "";
+            try {
+                var created_at = new Date(tweetJSON['create_at']);
+                tweetTime = created_at.toTimeString() + "\\t" + created_at.toDateString();
```
baiqiushi: From the screenshot, it seems this \\t doesn't make any effect there. Try another way to fill the space, e.g. html tag for whitespace is "&nbsp"

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -253,20 +256,21 @@ angular.module('cloudberry.map')
...
-              $scope.currentMarker.bindPopup($scope.popUpTweet).openPopup();
-            }).
-            error(function() {
-              var tweetContent = "Sorry! It seems the tweet with that ID has been deleted by the author.@_@";
+
+            //send the query to cloudberry.
+            var passID = "" + $scope.pointIDs[i];
+            cloudberry.pinMapOneTweetQuery(passID);
+
+            //receives the result and update content of tweet.
```
baiqiushi: For this part, I think the logic is not correct, please pay attention to the line 330-335 in this file.
The $watch function should accept 2 parameters: first is a function who returns the variable that we need to watch for its update, second is a function that will be called when the first variable is updated. Try to mimic that logic.

examples/twittermap/web/conf/application.conf
```diff
@@ -77,8 +77,9 @@ bounded-mailbox {
```
baiqiushi: I think this file should not be updated for this PR.

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -253,20 +282,21 @@ angular.module('cloudberry.map')
...
-              $scope.popUpTweet.setContent(tweetContent);
-              $scope.currentMarker.bindPopup($scope.popUpTweet).openPopup();
-            });
+
+            //send the query to cloudberry.
+            var passID = "" + $scope.pointIDs[i];
+            cloudberry.pinMapOneTweetQuery(passID);
+
+            //monitors and receives the result with updating content of each pin tweet.
+              $scope.$watch(function () {
```
baiqiushi: Here, the $watch function should be called outside of the $scope.map.on function, because currently, everytime the mouse hang over a point, it will register the $watch variable and function once.

chenlica:
What's the reason of the "wrong id" bug @ChankiWu ?

examples/twittermap/web/conf/application.conf
ChankiWu: Yep, we have fixed this I think.

ChankiWu:
@chenlica To be simple, the original logic of jianfeng's code is BigDecimal->long->double->long, so during this process, double->long may cause some errors. Now due to jianfeng's great help, we change the logic to BigDecimal->long, so it is more concise and simple with no errors during conversion of JSNumber. Will explain in our meeting this week again if you believe it is necessary. Thanks for caring !

ChankiWu:
Now the pinmap talks to cloudberry works pretty fine with every single point showing its corresponding tweet.

Now screenshot of current version:

examples/twittermap/web/conf/application.conf
waans11: Let's not change the application.conf file.

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -88,13 +88,13 @@ angular.module('cloudberry.map')
       }
 
       $scope.loadGeoJsonFiles(onEachFeature);
-      
+
```
waans11: Why is this empty line required?

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,95 @@ angular.module('cloudberry.map')
...
-        function translateOembedTweet(tweetJSON) {
-          var userName = "";
-          try {
-            userName = tweetJSON.author_name;
-          }
-          catch (e){
-            console.log("author_name missing in this Tweet.:" + e.message);
-          }
+        //shows each point's info in Front-end.
+        function translateTweetdatatoShow(tweetJSON) {
```
waans11: translateTweetDataToShow

examples/twittermap/web/public/javascripts/map/pointmap.js
```diff
@@ -158,66 +158,95 @@ angular.module('cloudberry.map')
...
+            var tweetid = "";
+            try {
+                tweetid = tweetJSON["id"];
+            }
+            catch (e){
+                console.log("tweetid missing in this Tweet. :" + e.message);
+            }
```
waans11: These console.log() are really necessary?

examples/twittermap/web/public/javascripts/map/pointmap.js
ChankiWu: It is because Qiushi provides me with new sample data, so we have change the application.conf file to make it work. I think the empty line should be the format issue. Let me see if I can adjust it.

examples/twittermap/web/public/javascripts/map/pointmap.js
ChankiWu: Technically, we don't need this try catch and console.log, because now I could get every point's tweet. I keep this because Qiushi used to write these try-catch structure in his function, and i think it is a good habit to code.

examples/twittermap/web/public/javascripts/map/pointmap.js
ChankiWu: Sorry for that. Not very precise for Camel-Case rule.

examples/twittermap/web/public/javascripts/map/pointmap.js
waans11: Please remove the application.conf change in your change. You can do this by overwriting the current one to your commit.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -11,6 +11,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       sentimentUpperBound: 4,
       cacheThreshold: parseInt(config.cacheThreshold),
       querySliceMills: parseInt(config.querySliceMills),
+      pinMapOneTweetResult: null,
```
waans11: Let's change the name to the following to make it more clear.

pinMapOneTweetLookupResult

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
             // unrecognized map type
             break;
         }
+      },
+
+      pinMapOneTweetQuery: function(pinid){
```
waans11: 1. Name change: pinMapOneTweetLookUpQuery
2. Please put a comment about this query. 
// This query retrieves the information about a specific tweet that the user selected.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+
+          var setpinfilter = [{
```
waans11: Change the name more to contain more info.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+              field: "id",
+              relation: "in",
+              values: pinid
+          }];
+
+          //sql
```
waans11: Let's remove this comment. It's not SQL.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+          var pinDBquery = (JSON.stringify({
+              dataset:"twitter.ds_tweet",
+              filter: setpinfilter,
+              select:{
+                  order: ["-create_at"],
+                  limit: defaultSamplingSize,
```
waans11: Actually, I am not sure why order and limit are required. This query is supposed to return only one result. Right?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                  offset: 0,
+                  field: ["id","text","user.id","create_at","user.name","user.profile_image_url"]
+              },
+              transform:{
+                  wrap:{
+                      id:"pinMapOneTweetResult",
```
waans11: pinMapOneTweetLookupResult

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                      category:"pinMapOneTweetResult"
```
waans11: pinMapOneTweetLookupResult

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -434,6 +435,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                  }
+              }
+          }));
+
+          //changes the id's type from string to int64.
+          pinDBquery = pinDBquery.replace(/"(\d+)"/,"[$1]");
```
waans11: This just removes double quotes?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -485,6 +518,9 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
           case "totalCount":
             cloudberryService.commonTotalCount = result.value[0][0].count;
             break;
+          case "pinMapOneTweetResult":
```
waans11: Here, too.

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: Sure. Thanks for your patience. Will change !

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: Yep, it could be deleted I think.

examples/twittermap/web/public/javascripts/common/services.js
waans11: Thanks. pinMapOneTweetLookUpResult

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: Yep, I tried many ways and this one works fine. Plus, we need to make sure it is int so we must remove double quotes.

examples/twittermap/web/public/javascripts/common/services.js
ChankiWu: I test and find these fields are all required otherwise the cloudberry will give errors. So although we may not need to specify it, we have to do write that.

ChankiWu: So these fields can not be deleted.

ChankiWu:
@baiqiushi OK. Previously Taewoo told me that application.log shouldn't change so I just rolled back that commit. Anyway, will change the time range. Thanks !

PR #497, opened 2018-03-20
new objects->static objects

liming-thu:


examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -16,22 +16,29 @@
     public static final SimpleDateFormat ADMDateFormat = new SimpleDateFormat("yyyy-MM-dd");
     public static final SimpleDateFormat ADMTimeFormat = new SimpleDateFormat("HH:mm:ss.SSS");
 
+    public static StringBuilder sbConstructor=new StringBuilder();
```
JavierJia: This static StringBuilder is stateful, thus it is error-prone if used in multiple method.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -16,22 +16,29 @@
...
     public static String mkADMConstructor(String constructor, String content) {
-        return constructor + "(\"" + content + "\")";
+        sbConstructor.delete(0,sbConstructor.length());
```
JavierJia: this made the `mkADMConstructor` non-functional. Why not just use the local variable?

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -40,7 +47,9 @@ public static String mkDateTimeConstructor(Date jdate) {
         return "datetime(\"" + ADMDateFormat.format(jdate) + "T" + ADMTimeFormat.format(jdate) + "\")";
     }
 
-    public static Rectangle coordinates2Rectangle(GeoLocation[][] boundingBoxCoordinates){
+    public static StringBuilder sbDate = new StringBuilder();
```
JavierJia: same concern here.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -74,19 +83,19 @@ public static Rectangle coordinates2Rectangle(GeoLocation[][] boundingBoxCoordin
         }
         return new Rectangle(swLog, swLat, neLog, neLat);
     }
-
+    public static StringBuilder sbRectangleConstructor=new StringBuilder();
```
JavierJia: same here.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -97,43 +106,46 @@ public static void keyValueToSbWithComma(StringBuilder sb, String key, String va
         keyValueToSb(sb, key, val);
         sb.append(",");
     }
-
+    public static StringBuilder sbPoint=new StringBuilder();
```
JavierJia: same here.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -97,43 +106,46 @@ public static void keyValueToSbWithComma(StringBuilder sb, String key, String va
...
-        return "point(\"" + geoLocation.getLongitude() + "," + geoLocation.getLatitude() + "\")";
+        sbPoint.delete(0,sbPoint.length());
+        sbPoint.append("point(\"" ).append(geoLocation.getLongitude()).append(",").append(geoLocation.getLatitude()).append("\")");
+        return  sbPoint.toString();
     }
 
     public static String mkPoint(String lng, String lat) {
         return "point(\"" + lng + "," + lat + "\")";
     }
+    public static StringBuilder sbStringSetHash=new StringBuilder();
```
JavierJia: ditto.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
```diff
@@ -97,43 +106,46 @@ public static void keyValueToSbWithComma(StringBuilder sb, String key, String va
...
-            sb.append(mkQuote(hashtagEntities[i].getText()));
+            sbStringSetHash.append(mkQuote(hashtagEntities[i].getText()));
         }
-        sb.append("}}");
-        return sb.toString();
+        sbStringSetHash.append("}}");
+        return sbStringSetHash.toString();
     }
-
+    public static StringBuilder sbStringSetUserMention=new StringBuilder();
```
JavierJia: ditto.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/Place.java
```diff
@@ -11,12 +10,12 @@
     public static String NAME = "name";
     public static String PLACE_TYPE = "place_type";
     public static String BOUNDING_BOX = "bounding_box";
-
+    public static StringBuilder placeSB = new StringBuilder();
```
JavierJia: ditto.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/Tweet.java
```diff
@@ -21,73 +23,73 @@
     public static String USER = "user";
     public static String PLACE = "place";
 
-    public static String toADM(Status status, USGeoGnosis gnosis, boolean requireGeoField) throws UnknownPlaceException{
+    public static StringBuilder admSB = new StringBuilder();
```
JavierJia: ditto

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/Tweet.java
```diff
@@ -21,73 +23,73 @@
...
+        ADM.keyValueToSb(admSB, USER, User.toADM(status.getUser()));
+        admSB.append("}");
+        return admSB.toString();
     }
-
-    public static String geoTag(Status status, USGeoGnosis gnosis, boolean requireGeoField) throws UnknownPlaceException{
-        StringBuilder sb = new StringBuilder();
-        if (textMatchPlace(sb, status, gnosis)) {
-            return sb.toString();
+    public static StringBuilder sbGeoTag=new StringBuilder();
```
JavierJia: ditto.

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/adm/ADM.java
baiqiushi: @JavierJia  Hi, the background here is our AsterixDB is not stable, and re-ingestion process has been repeated for several times, each time consuming 3-4 days. Therefore, @chenlica asks @liming-thu to look into how to accelerate this process.

For your concern, do you mean, if he uses this static stringbuilder, it might mess up the data with incorrect output? If so, I think we need to think about it more carefully and might need some more semantic tests. Thanks!

@liming-thu  Could you provide some abstract description of what you have found about this issue, e.g. which part of the code occupies how much percent of the total time of the geoTag process? So that people can understand the motivation here. And besides, a PR needs some description any way. Please refer to previous PRs, about how to write a description. Thanks!

JavierJia: It will mess up for sure. One simple counter argument is that how it work in the multi threading case? 

If the context is to speed up the geotagging process, one direction I can think of is to run the command in parallel, either multi-threading or multi-process. It would improve the performance x-times faster depending the degree of parallelism. 

Yes, I agree that we can also optimize the code, which may achieve x% performance gain. If we deciding to going to this direction, I believe that we've already used some tool (e.g., [Yourkit](https://www.yourkit.com/)) to analyze the bottleneck. Could anyone share the number of the analysis?

Even it indeed the case that the builder can save a lot of time, we should pass the `builder` into the function as a parameter instead of creating a global static variable.

chenlica:
Suggestion: we do this PR together with the task of multi-threading.  We will revisit this topic after doing multi-threading experiments.

PR #499, opened 2018-03-29
Aggressive drum reporter

baiqiushi:
* Previously, every time when it reaches deadline for reporting result, the `reporter` only dequeue one partial result of one mini query to return to the client.
* Currently, every time when it reaches deadline for reporting result, the `reporter` will empty the partial result queue and return them all.


waans11:
Could you provide an answer to my question?

baiqiushi:
@waans11  We need to talk about this later F2F with @chenlica.

JavierJia:
Don't know if you already made the decision. It feels more safe to keep at least one at stock IMO. Anyway it should be based on the user experience on the Twittermap.

baiqiushi:
**Abandon Reason**
Finally, we found that this queue issue is not the real reason of the Drum's conservativeness.

**Future reference**
If we need to make this queue issue to be more aggressive, we should modify that enqueue logic part instead of this part. We could dequeue (n-1) elements of the queue before enqueue a new partial result.

PR #501, opened 2018-04-03
Disable create views in RESTSolver and Tune alpha to be more aggressive

baiqiushi:
**Disable create views in RESTSolver**
Currently, when Twittermap sends 2 concurrent queries to Cloudberry, one being batch query handled by ProgressiveSolver with slicing and the other being single query handled by RESTSolver without slicing. The latter will be finished very soon and create views in AsterixDB while ProgressiveSolver is still running and accessing the same part of data in AsterixDB, then the creating views operation will compete resources with ongoing mini queries to slow them down. This PR will `disable` this `views creation` operation in `RESTSolver` when non-slicing queries finish to avoid this problem.

**Tune alpha to be more aggressive**
Our observation shows that current Drum behavior in production system with 1.2 billion tweets is too `conservative` on uniformly distributed keywords, therefore, this PR also modifies the `Alpha` parameter of `Drum` from `1.0` to `0.00001` to be much more `aggressive`, so that the power of AsterixDB can be fully utilized.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/RESTSolver.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/RESTSolver.scala
@@ -26,6 +26,7 @@ class RESTSolver(val dataManager: ActorRef,
       val futureResult = Future.traverse(queries)(q => solveAQuery(q)).map(JsArray.apply)
       futureResult.map(result => (queries, result)).foreach { case (qs, r) =>
         out ! transform.transform(r)
+        //Disabled this suggest views to avoid competing resources in DB with ongoing ProgressiveSolver.
         //qs.foreach(suggestViews)
       }
   }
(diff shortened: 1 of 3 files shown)


chenlica:
LGTM

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/RESTSolver.scala
```diff
@@ -26,7 +26,8 @@ class RESTSolver(val dataManager: ActorRef,
       val futureResult = Future.traverse(queries)(q => solveAQuery(q)).map(JsArray.apply)
       futureResult.map(result => (queries, result)).foreach { case (qs, r) =>
         out ! transform.transform(r)
-        qs.foreach(suggestViews)
+        //Disabled this suggest views to avoid competing resources in DB with ongoing ProgressiveSolver.
```
JavierJia: I would suggest adding a TODO or FIXME so that it will be a little more noticeable.

baiqiushi: @JavierJia  Great! Thanks!

PR #502, opened 2018-04-06
Fix bug: conf berry.query.gap is never used in Drum

baiqiushi:
**Bug**
Previously, the `ProgressiveSolver` will `initialize` a `Drum` instance with the parameter `minRange` assigned value from `berry.firstquery.gap` in conf file. But the `berry.query.gap` parameter never used.

**Solution**
In `ProgressiveSolver`, load the `berry.query.gap` value from conf file and initialize `Drum` instance with `minRange` assigned by that value.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
@@ -58,12 +58,12 @@ class ProgressiveSolver(val dataManager: ActorRef,
       val boundary = new TInterval(min, max)
 
       val initialDuration = config.FirstQueryTimeGap
-      val minDuration = config.MinTimeGap
+      val minimumDuration = config.MinTimeGap
       val interval = calculateFirst(boundary, initialDuration)
       val queryGroup = QueryGroup(ts, queryInfos, request.postTransform)
       val initResult = Seq.fill(queryInfos.size)(JsArray())
       issueQueryGroup(interval, queryGroup)
-      val drumEstimator = new Drum(boundary.toDuration.getStandardHours.toInt, alpha = 0.00001, minDuration.toHours.toInt)
+      val drumEstimator = new Drum(boundary.toDuration.getStandardHours.toInt, alpha = 0.00001, minimumDuration.toHours.toInt)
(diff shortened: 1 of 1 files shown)


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
```diff
@@ -58,11 +58,12 @@ class ProgressiveSolver(val dataManager: ActorRef,
       val boundary = new TInterval(min, max)
 
       val initialDuration = config.FirstQueryTimeGap
+      val minDuration = config.MinTimeGap
```
waans11: minDuration is confusing. (duration in minutes OR minimum duration?) We know it's minimum duration. So, let's call it minimumDuration.

baiqiushi: Sure.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
```diff
@@ -58,11 +58,12 @@ class ProgressiveSolver(val dataManager: ActorRef,
       val boundary = new TInterval(min, max)
 
       val initialDuration = config.FirstQueryTimeGap
+      val minimumDuration = config.MinTimeGap
```
chenlica: What if the user didn't provide this value in the config file?  If so, should we assign a default value?

waans11: Normally the front-end assigns a default value for parameters. @baiqiushi: can you check for this parameter?

baiqiushi: Yes. We have default value assigned when the `config` object is initiated in `zion/src/main/scala/edu.uci.ics.cloudberry.zion/common/Config` file:
`val MinTimeGap = config.getString("berry.query.gap").map(parseTimePair).getOrElse(1 day)`.
So, for this case, no problem.

chenlica:
LGTM

PR #505, opened 2018-04-14
Fixed a bug when the maptype is changed

DeyuChen:
Fixed a minor bug when the maptype is switched to pinmap or heatmap from countmap, some information only related to countmap remains visible on the map.

Original screenshot
screenshot from 2018-04-14 16-35-27

Fixed screenshot
screenshot from 2018-04-14 16-34-50


examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -62,11 +62,19 @@ angular.module('cloudberry.map')
...
+        if (ctrlClass) {
+          ctrlClass.hide();
+        }
+      }
 
       // remove CountMap controls
       removeMapControl('legend');
       removeMapControl('normalize');
       removeMapControl('sentiment');
+      hideMapControl('info');
```
waans11: What's the difference between removeMapControl and hideMapControl? Why can't we use the currently existing method? :-)

DeyuChen: hideMapControl simply makes the div invisible by setting the hide option. It is easier to make the div reappear again later. The original removeMapControl method deletes the div entirely and create them later. Because the other 3 map controls are added in a different place with 'info', the code cannot be used in the first place, so I figured I can use a different approach to hide the div.

DeyuChen: I tried create/remove instead of show/hide for 'info' but there are some problems. Maybe we can discuss it face to face whenever you are available?

baiqiushi: @waans11 , could you approve this PR and merge it?

examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -141,7 +142,14 @@ angular.module('cloudberry.map')
       info.options = {
         position: 'topleft'
       };
-      $scope.controls.custom.push(info);
+      if ($scope.map){
+        $timeout(function(){
```
waans11: What's the meaning of this timeout? I see we are adding the info here.

DeyuChen: I added the timeout to avoid some weird behavior as follow:

mapbug

Every time I add the div back, it looks like this for a second before the correct text is shown. I think it is because it takes some time for the compile (line 138) to resolve the data. Right now I haven't found a better way to solve this problem.

PR #519, opened 2018-06-04
[issue#513] timeseries caching

ilvry:
**TimeSeries Caching**:
Preload all data required to generate timeseries for each county (or other selected level) when a keyword is newly inserted and cached the result in module timeseriescache. The cached data can be used to generate (selected level's) map result/individual count (by this.putValues, this.getValues) and timeseries for valid time intervals (by this.getTimeSeriesValues in service TimeSeriesCache).

**Goals**: 
1. Change byTimeSeriesRequest query to receive polygonId and histogram key-value pairs, return <region ID, [date:count, date2:count2, …]>.
Merge [date:count, …] to generate the current timeseries result.
2. Introduce timeseriescache module.
3. The front-end logic can utilize this timeseriescache module.

This PR will forcus on (1).

### DIFF
--- a/examples/twittermap/web/public/javascripts/common/services.js
+++ b/examples/twittermap/web/public/javascripts/common/services.js
@@ -123,42 +123,6 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       return filter;
     }
 
-    function byTimeSeriesRequest(parameters, geoIds) {
-        return {
-        dataset: parameters.dataset,
-        filter: getFilter(parameters, defaultNonSamplingDayRange, geoIds),
-        group: {
-          by: [{
-            field: "geo",
-            apply: {
-              name: "level",
-              args: {
-                level: parameters.geoLevel,
-              }
-            },
-            as: parameters.geoLevel
-          },
-          {
-            field: "create_at",
-            apply: {
-              name: "interval",
-              args: {
-                unit: parameters.timeBin
-              }
-            },
-            as: parameters.timeBin
-          }],
-          aggregate: [{
-            field: "*",
-            apply: {
-              name: "count"
-            },
-            as: "count"
-          }]
-        }
-      };
-    }
-
     function byGeoRequest(parameters, geoIds) {
       if (cloudberryConfig.sentimentEnabled) {
         return {
@@ -235,12 +199,22 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     }
 
-    function byTimeRequest(parameters) {
-      return {
+    function byTimeRequest(parameters, geoIds) {
+        return {
         dataset: parameters.dataset,
-        filter: getFilter(parameters, defaultNonSamplingDayRange, parameters.geoIds),
+        filter: getFilter(parameters, defaultNonSamplingDayRange, geoIds),
         group: {
           by: [{
+            field: "geo",
+            apply: {
+              name: "level",
+              args: {
+                level: parameters.geoLevel,
+              }
+            },
+            as: parameters.geoLevel
+          },
+          {
             field: "create_at",
             apply: {
               name: "interval",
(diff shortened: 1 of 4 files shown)


ilvry:
[commit 99b0d3f] add JSON request byTimeSeriesRequest
[future commit] will be send as preloadRequest by the Common service module, MapResultCache.putValues when ws.onmessage.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -90,7 +90,9 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
     }
 
-    function getFilter(parameters, maxDay, geoIds) {
+    const GetAllInLevel = 1;
```
waans11: Should be a boolean value? And why do we need this?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -122,6 +124,46 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       }
       return filter;
     }
+    
+    // Request all data required by timeseriescache.
```
waans11: Why do we need to have a new TimeSeries Request query?

ilvry: By far, byTimeRequest or byGeoRequest only give the result of timeseries ({day, count}) or map result ({county, count, population}). If we want to use this cache to generate both, we can have {county, day, count, population} by sending this one request. Also, no new request needs to be sent when a time interval is selected and we can use it to display timeseries histogram for each selected level (e.x.county) region.

examples/twittermap/web/public/javascripts/common/services.js
ilvry: Yes, it should be boolean value.
This parameter allows me to request all units of the same selected level (e.x.county), without listing all countyIds.
I have a question about this choice of number in the original code: geoIds.length <= 2000?

ilvry: Is there a specific reason behind "2000" (an approximation of time spending)?

waans11: I think it's about the size of "request". If it's bigger, the byte size becomes bigger, and the system was not able to send/receive it correctly because of the buffer size.

ilvry: Understood, thanks.

ilvry:
[commit 7a4116a] part#1
Change byTimeSeriesRequest query to receive polygonId and histogram key-value pairs, return <region ID, [date:count, date2:count2, …]>.
Merge [date:count, …] to generate the current timeseries result.

Comments in the following format: "//temp:lineNumber,future location" will be moved due to the addition of timeseriescache module.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -299,9 +309,21 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
               }
             }));
 
+            var geoIdsNotInTimeSeriesCache = parameters.geoIds;
```
waans11: This variable doesn't need to be present in the step 1.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -299,9 +309,21 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+            // for the time histogram
+            var timeSeriesRequest = (JSON.stringify({
```
waans11: Why do we need this variable? Can we use it like the original byTimeRequest()?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -357,6 +379,7 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
                 cloudberryService.parameters.geoLevel);
 
               ws.send(sampleJson);
+              ws.send(timeSeriesRequest);
```
waans11: Again, this request should be in a batch, not a separate request.

PR #520, opened 2018-06-04
Add timeout of 5 mins for curl command

baiqiushi:


chenlica:
Agreed.  If the overhead of each query is low, we can use a shorter period.

baiqiushi:
Of course. Adjustments made.

Issue #521, opened 2018-06-08
TwitterMap's total count is more than actual number in DB


ChankiWu:
According to Qiushi's observation and our tests, the total count in TwitterMap is far beyond the actual cardinality of tweets in AsterixDB. That is to say, Cloudberry's Metadata is not correspond to tweets' number in AsterixDB, because the tweets showed in TwitterMap is provided by Cloudberry.

PR #522, opened 2018-06-08
[Issue #521] Fix the issue of Tweet's total count in Cloudberry.

ChankiWu:
Symptom: In TwitterMap, the number shown at the lower left corner is not corresponding to the actual number in AsterixDB. 

Reason: Cloudberry keeps a Metadata of DB. TwitterMap sends a request to Cloudberry to update its total count of tweets every second. Then, Cloudberry sends a request to AsterixDB to check if a "newCount" in DB has been fed every 4 hours. If so, Cloudberry will add newCount to the lastCount to produce a "new" count and give it back to TwitterMap. However, the "dataInterval" of lastCount is wrong because it didn't record the newCount.till time (which is the lastTime Cloudberry sent a request, only in memory) in MetaData, so when we restart cloudberry, the timestamp of the count is incorrect.


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -57,7 +57,7 @@ class OriginalDataAgent(val dataSetInfo: DataSetInfo,
   override protected def maintenanceWork: Receive = {
     case newCount: Cardinality =>
       lastCount.reset(lastCount.from, newCount.till, lastCount.count + newCount.count, newCount.ratePerSecond)
-      context.parent ! NewStats(dbName, newCount.count)
+      context.parent ! NewStats(dbName, newCount.count, new Interval(lastCount.from,newCount.till))
```
JavierJia: `lastCount.from,newCount.till` -> `lastCount.from, newCount.till`

chenlica:
Is there any way to test this fix?  Can we add a test for this bug fix?

JavierJia:
There are some unit tests using the `NewStats` object. The error message said
> [error] /home/travis/build/ISG-ICS/cloudberry/cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/DataStoreManagerTest.scala:190: not enough arguments for method apply: (dbName: String, additionalCount: Long, dataInterval: org.joda.time.Interval)edu.uci.ics.cloudberry.zion.actor.OriginalDataAgent.NewStats in object NewStats.

You should fix that test. I think you can run the cmd `sbt 'project zion' test ` to verify if the test has passed.

cloudberry/zion/src/test/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgentTest.scala
```diff
@@ -126,12 +126,16 @@ class OriginalDataAgentTest extends Specification with Mockito {
       val cardinality = Cardinality(initialMaxTime, dateNow, additionalCount)
 
       val agent = TestActorRef(OriginalDataAgent.props(dataSetInfo, mockQueryParser, mockConn, Config.Default), parent.ref, "child")
-      parent.expectMsg(NewStats(dataSetInfo.name, initialCount))
+      var newTestStates = parent.receiveOne(3.seconds).asInstanceOf[NewStats]
+      assert( newTestStates.additionalCount == initialCount )
+      assert( newTestStates.dbName == dataSetInfo.name )
```
JavierJia: hmm, skipping the time interval check is not a correct way of fixing the test. any clue why the ending time of the interval is different from dateNow?

ChankiWu: Oh, yep. In fact, Qiushi and I was planning to add the interval check, but the initial count and additional count's interval are the same and a fixed value, which means it makes not much sense to test the interval. The difference is very small and about 3ms, because the place to assign ending time is in different files.

JavierJia:
Could you point the place that sets the time?
If that is the expected behavior, we can test the logic by asserting the
delta of two timestamps is less than a certain number, say 1s, other than
leave it unchecked, right?

ChankiWu:
OK. Based on Jianfeng's suggestion, I added the dateInterval test referring to the test case. After several tests, I set the deviation to 5s.

chenlica:
@ChankiWu @JavierJia @baiqiushi : can we finish this PR completely before @ChankiWu leaves for the summer?  Thanks.

PR #525, opened 2018-06-21
[issue#513] Time-series caching part#1

ilvry:
Part#1: Modify byTimeRequest and merge query results to generate timeseries.
- `common/services.js`
(i) `byTimeRequest(parameter, geoIds)` returns [<geoId:time:count>, ...]
(ii) `handleByTimeRequest(parameters, geoIds)` handles `heatmap`'s and `pointsmap`'s byTimeRequest.
(iii) `getTimeSeriesValues(byTimeRequestResult)`: change byTimeRequest results [<geoId:time:count>, ...] to HashMap{<geoId,[<date:count>, ...]>, ...}, then return current geoId's time-series data [<date:count>, ...] 

________
Part#2: Introduce timeseriescache module, which includes the HashMap store and merge function.
- `app.js` and `main.scala.html`: add time-series cache module.
- `common/services.js`: calls `timeseriescache/services.js` in similar ways to mapresultcache
(i) `getGeoIdsNotInCache()` returns geoIdsNotInTimeSeriesCache
(ii) `getTimeSeriesValues()` returns in-cache time-series data
(iii) `putTimeSeriesValues()` returns newly requested time-series data

________
Part#3: Utilize the time-series cache module in front-end logic.
- `common/services.js`


examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -568,17 +592,56 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
           case "sample":
             cloudberryService.commonTweetResult = result.value[0];
             break;
+          case "timeSeriesRequest":
+            var prefix = Object.freeze({
```
baiqiushi: As far as I know, the stateID, countyID and cityID will not overlap with each other @waans11 can confirm that? it might not be necessary for this prefix? Then might save some space in case data size increases.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -568,17 +592,56 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                
+            var timeSeriesStore = new HashMap();
+            var geoIdSet = new Set(cloudberryService.parameters.geoIds);
+            for (var i = 0; i < result.value[0].length; i++) {
+              var currVal = result.value[0][i];
+              var geoIds = prefix[cloudberryService.parameters.geoLevel] + currVal[cloudberryService.parameters.geoLevel];
+              var values = timeSeriesStore.get(geoIds);
+              // First updates the store with geoIds that have results.
+              if (values !== undefined && values !== INVALID_VALUE) { // when one geoIds has more than one value
+                values.push({day:currVal["day"], count:currVal["count"]});
```
baiqiushi: As we discussed, the overhead of communication on entire String of `day` attribute for the bins might be large when number of records returned reach to thousands. If we can use offset # of days after the `startDate` as bin index, it might be more efficient.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -568,17 +592,56 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+              }
+            }
+
+            cloudberryService.commonTimeSeriesResult = timeSeriesResultArray;
+            break;
           // Complete cache hit case
           case "batchWithoutGeoRequest":
             if(angular.isArray(result.value)) {
-              cloudberryService.commonTimeSeriesResult = result.value[0];
+              //cloudberryService.commonTimeSeriesResult = result.value[0];
```
baiqiushi: Let's remove useless code instead of commenting it.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -568,17 +592,56 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
               cloudberryService.commonHashTagResult = result.value[1];
             }
             break;
           // Partial map result cache hit or complete cache miss case
           case "batchWithPartialGeoRequest":
             if(angular.isArray(result.value)) {
-              cloudberryService.commonTimeSeriesResult = result.value[0];
+              //cloudberryService.commonTimeSeriesResult = result.value[0];
```
baiqiushi: Same as above.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -225,6 +235,31 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
       };
     }
 
+    function handleByTimeRequest(parameters, geoIds) {
```
baiqiushi: In this function, we don't need a batch request at all. A single query can also specify an `option` attribute:
```
{
          dataset: parameters.dataset,
          filter: getFilter(parameters, defaultNonSamplingDayRange, parameters.geoIds),
          group: {
            by: [{
              field: "create_at",
              apply: {
                name: "interval",
                args: {
                  unit: parameters.timeBin
                }
              },
              as: parameters.timeBin
            }],
            aggregate: [{
              field: "*",
              apply: {
                name: "count"
              },
              as: "count"
            }]
          },
          option: {
            sliceMillis: cloudberryConfig.querySliceMills
          }
        }
```
And for in different cases, you may need to add this attribute dynamically, you can always manipulate an object in Javascript like this:
```
// Suppose query is previously constructed JSON object
// Add "transform" attribute to the query JSON
        query["transform"] = {
          wrap: {
            id: id ? id : category,
            category: category
          }
        };
```

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -279,6 +314,9 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
         if (ws.readyState !== ws.OPEN || typeof(parameters.keywords) === "undefined" || parameters.keywords == null || parameters.keywords.length == 0)
           return;
 
+        // for the time-series histogram
+        handleByTimeRequest(parameters, parameters.geoIds);
```
baiqiushi: We should not call this `handleByTimeRequest` function in any conditions, since in the following switch-case blocks, when `countmap` you also send `byTimeRequest` in the batch query. Or otherwise, you may need to get rid of those `byTimeRequest` queries in the batches.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -568,17 +536,51 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+                if (values !== undefined && values !== INVALID_VALUE) { // when one geoIds has more than one value
+                  values.push({day:currVal["day"], count:currVal["count"]});
+                  timeSeriesStore.set(geoIds, values);
+                  geoIdSet.delete(currVal[cloudberryService.parameters.geoLevel]);
+                } else { // first value of current geoId
+                  timeSeriesStore.set(geoIds, [{day:currVal["day"], count:currVal["count"]}]);
+                  geoIdSet.delete(currVal[cloudberryService.parameters.geoLevel]);
+                }
+                // Mark other results as checked: these are geoIds with no results
+                geoIdSet.forEach(function (value) {
```
baiqiushi: This cleanup step should be outside the for loop.

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -224,6 +234,24 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
         }
       };
     }
+    
+    function handleByTimeRequest(parameters, categoryName) {
```
waans11: Can you put a reason as a comment why the following check is needed?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -571,14 +531,77 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+              var timeSeriesResultArray = [];
+              for (var j = 0; j < cloudberryService.parameters.geoIds.length; j++) {
+                var value = timeSeriesStore.get(cloudberryService.parameters.geoIds[j]);
+                if (value !== undefined && value !== INVALID_VALUE) {
+                  timeSeriesResultArray = timeSeriesResultArray.concat(value);
+                }
+              }
+
+              cloudberryService.commonTimeSeriesResult = timeSeriesResultArray;
+
```
waans11: lines 534-565 and 572-604 look duplicated. Can we refactor them as a function?

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -601,7 +624,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
+              var timeSeriesResultArray = [];
+              for (var j = 0; j < cloudberryService.parameters.geoIds.length; j++) {
+                var value = timeSeriesStore.get(cloudberryService.parameters.geoIds[j]);
+                if (value !== undefined && value !== INVALID_VALUE) {
+                  timeSeriesResultArray = timeSeriesResultArray.concat(value);
+                }
+              }
+
+              cloudberryService.commonTimeSeriesResult = timeSeriesResultArray;
```
waans11: Here, too...

examples/twittermap/web/public/javascripts/common/services.js
```diff
@@ -612,7 +666,38 @@ angular.module('cloudberry.common', ['cloudberry.mapresultcache'])
...
```
waans11: Here, too...

examples/twittermap/web/public/javascripts/common/services.js
baiqiushi: @ilvry you only call this `handleByTimeRequest` function in `heatmap` and `pinmap` cases, right? Then no need to check that again, I think.

PR #527, opened 2018-06-26
Fix bug when incoming tweet out of US, geotag throws exception

baiqiushi:
## Symptom ##
When incoming tweet from twitter.com is out of US, geotag cannot find the cityID by its city name or city code, geotag throws an exception for `coordinates` is null because the tweet has no `coordinates` attribute.

## Solution ##
Add an `if` check for this kind of situation


baiqiushi:
Some synchronization problem on creating this PR. Abandoned.

PR #528, opened 2018-06-26
Add null check for coordinates

baiqiushi:
Temporary PR for this bug.


baiqiushi:
Still there's other bug for this geotag module, needs to be looked into later by @liming-thu

PR #537, opened 2018-07-03
Frontend refactor - Module management

baiqiushi:
Introduce module-manager who provides API functions for modules to publish and subscribe events.


examples/twittermap/web/public/javascripts/common/module-manager.js
```diff
@@ -0,0 +1,81 @@
...
+        if (!(eventName in this.EVENT)) {
+          return false;
+        }
+
+        if (eventHandler instanceof Function) {
+          if (eventName in this.eventsListeners) {
+            this.eventsListeners[eventName].add(eventHandler);
+          }
+          else {
+            this.eventsListeners[eventName] = new Set([eventHandler]);
```
waans11: Is [eventHandler] correct? Should it be Set(eventHandler)?

examples/twittermap/web/public/javascripts/common/module-manager.js
```diff
@@ -0,0 +1,105 @@
+angular.module("cloudberry.common")
```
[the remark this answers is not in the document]
waans11: And let's add examples regarding subscribe, unsubscribe, and publish.

examples/twittermap/web/public/javascripts/common/module-manager.js
```diff
@@ -0,0 +1,81 @@
...
+    var moduleManager = {
+
+      // Built-in events
+      EVENT: {
+        CLICK_SEARCH_BUTTON: "clickSearchButtonEvent",
+        CHANGE_TIME_SERIES_RANGE: "changeTimeSeriesRangeEvent",
+        CHANGE_ZOOM_LEVEL: "changeZoomLevelEvent",
+        CHANGE_REGION_BY_DRAG: "changeRegionByDragEvent",
+        CHANGE_MAP_TYPE: "changeMapTypeEvent"
+      },
```
waans11: Can't we remove "click..." string?

examples/twittermap/web/public/javascripts/common/module-manager.js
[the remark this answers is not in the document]
baiqiushi: Sorry, previous push is not final, only for switching branch convenience.  Now all the comments and examples added.

examples/twittermap/web/public/javascripts/common/module-manager.js
baiqiushi: This is correct. https://www.geeksforgeeks.org/sets-in-javascript/
The parameter for construction of Set must be iterable object instead of single object.

examples/twittermap/web/public/javascripts/common/module-manager.js
baiqiushi: I tried, not able to remove them. And after some investigation, to implement an enumeration object is not trivial, so I gave up the idea to make it close, now this object is only for the sake of readability of code, without any constraint for users.

waans11: Got it. I just checked too. Thanks for the investigation. Can we just use a simple constant after the event type such as "CHANGE_MAP_TYPE: 4" since the string part isn't really necessary.

examples/twittermap/web/public/javascripts/common/module-manager.js
waans11: OK.

examples/twittermap/web/public/javascripts/common/module-manager.js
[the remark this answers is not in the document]
waans11: Good.

examples/twittermap/web/public/javascripts/common/module-manager.js
```diff
@@ -0,0 +1,126 @@
...
+       *   once this event happens, the eventHandler function will be called.
+       *
+       * @param eventName moduleManager.EVENT, event name
+       * @param eventHandler function, callback function(event): event is the content object of the event
+       * @returns {boolean} true: if subscribed successfully, false: otherwise
+       */
+      subscribeEvent(eventName, eventHandler) {
+
+        if (eventName in moduleManager.eventSet) {
+          return false;
```
waans11: Should be "not in"?

examples/twittermap/web/public/javascripts/common/module-manager.js
```diff
@@ -0,0 +1,140 @@
...
+       * check if eventName is valid
+       *
+       * @param eventName
+       * @returns {boolean} true: if valid, false: otherwise
+       */
+      isEventValid(eventName) {
+        var isValid = false;
+        Object.keys(moduleManager.EVENT).forEach(function(key) {
+          if (moduleManager.EVENT[key] === eventName) {
+            isValid = true;
```
waans11: add a break once it turns out to be true. We don't need to continue...

PR #539, opened 2018-07-09
Frontend refactor - Cloudberry client

baiqiushi:
Introduce cloudberry-client who provides ONE API function for modules to send JSON queries to Cloudberry and register the query results handler functions as well.

### DIFF
--- a/examples/twittermap/web/public/javascripts/common/cloudberry-client.js
+++ b/examples/twittermap/web/public/javascripts/common/cloudberry-client.js
@@ -57,9 +57,10 @@ angular.module("cloudberry.common")
         // The first time registering queryCategory
         if (!(queryCategory in cloudberryClient.queryToResultHandlerMap)) {
           cloudberryClient.queryToResultHandlerMap[queryCategory] = {};
-          cloudberryClient.queryToResultHandlerMap[queryCategory][queryID] = resultHandler;
-        } else if (!(queryID in cloudberryClient.queryToResultHandlerMap[queryCategory])) {
-          // Register resultHandler for this queryCategory and queryID
+        }
+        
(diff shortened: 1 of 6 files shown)


examples/twittermap/web/public/javascripts/common/cloudberry-client.js
```diff
@@ -0,0 +1,91 @@
...
+      }
+
+    };
+
+    ws.onmessage = function(event) {
+      $timeout(function() {
+        var result = JSONbig.parse(event.data);
+        var category = result.category;
+        var id = result.id;
+        if (category in cloudberryClient.queryToResultHandlerMap) {
```
waans11: This check is not necessary since we already know that the result from Cloudberry has at least "category". Checking whether id is null is enough.

examples/twittermap/web/public/javascripts/common/cloudberry-client.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    ws.onmessage = function(event) {
+      $timeout(function() {
+        var result = JSONbig.parse(event.data);
+        var category = result.category;
+        var id = result.id;
+        if (id in cloudberryClient.queryToResultHandlerMap[category]) {
+          cloudberryClient.queryToResultHandlerMap[category][id](id, result.value);
+        }
```
waans11: I think a better format is placing "} else {" in one row. Did you apply the format in IntelliJ?

baiqiushi: No, I did it manually. Actually I like this format personally, since `else` is aligned with `if` making it more neat.

PR #542, opened 2018-07-10
[issue#513] Time-series caching part#2

ilvry:
Part#1: PR#525 [issue#513] Time-series caching part#1
Modify byTimeRequest and merge query results to generate timeseries.

________
Part#2: Introduce timeseriescache module, which includes the HashMap store and merge function.
- `app.js` and `main.scala.html`: add time-series cache module.
- `common/services.js`: calls `timeseriescache/services.js` in similar ways to mapresultcache
(i) `getGeoIdsNotInCache()` returns geoIdsNotInTimeSeriesCache
(ii) `setCompleteTimeInterval()` sets cache time range after `byTimeRequest` query is executed completely
(iii) `getTimeSeriesValues()` returns in-cache time-series data
(iv) `putTimeSeriesValues()` returns newly requested time-series data

________
Part#3: Utilize the time-series cache module in front-end logic.
- `common/services.js`

### DIFF
--- a/examples/twittermap/web/public/javascripts/timeseriescache/services.js
+++ b/examples/twittermap/web/public/javascripts/timeseriescache/services.js
@@ -16,7 +16,7 @@ angular.module('cloudberry.timeseriescache', [])
         var endDate = new Date();
         // Deducts -1 to create a default value that can't exist
         endDate.setDate(endDate.getDate() - 1);
-        // Cached time interval, intermediate or complete byTimeRequest result.
+        // Cached time interval of complete byTimeRequest result.
         var cachedTimeRange = {
             start: new Date(),
             end: endDate
--- a/examples/twittermap/web/public/javascripts/timeseriescache/services.js
+++ b/examples/twittermap/web/public/javascripts/timeseriescache/services.js
@@ -50,6 +50,7 @@ angular.module('cloudberry.timeseriescache', [])
                     geoIdsNotInCache.push(geoIds[i]);
                 }
             }
+
             return geoIdsNotInCache;
         };
 
--- a/examples/twittermap/web/public/javascripts/timeseriescache/services.js
+++ b/examples/twittermap/web/public/javascripts/timeseriescache/services.js
@@ -21,11 +21,6 @@ angular.module('cloudberry.timeseriescache', [])
             start: new Date(),
             end: endDate
         };
-        // Time interval after byTimeRequest query is executed completely.
-        var completeTimeRange = {
-            start: new Date(),
-            end: endDate
-        };
         const INVALID_VALUE = 0;
 
         /**
@@ -45,15 +40,11 @@ angular.module('cloudberry.timeseriescache', [])
                 timeseriesStore.clear();
                 currentKeywords = keywords.slice();
                 currentGeoLevel = geoLevel;
-                completeTimeRange.start = timeInterval.start;
-                completeTimeRange.end = timeInterval.end;
-                // Set default value which each result is >= start and <= end.
-                cachedTimeRange.start = timeInterval.end;
-                cachedTimeRange.end = timeInterval.start;
+                cachedTimeRange.start = timeInterval.start;
+                cachedTimeRange.end = timeInterval.end;
 
                 return geoIds;
             }
-
             for (var i = 0; i < geoIds.length; i++) {
                 if (!timeseriesStore.has(geoIds[i])) {
                     geoIdsNotInCache.push(geoIds[i]);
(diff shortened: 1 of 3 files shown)


examples/twittermap/web/public/javascripts/timeseriescache/services.js
```diff
@@ -0,0 +1,149 @@
+/*
+ * This module caches time-series histogram data. The common services module communicates with
+ * middle ware and the time-series cache.
+ */
+'use strict';
```
waans11: What's this "strict"?

examples/twittermap/web/public/javascripts/timeseriescache/services.js
```diff
@@ -0,0 +1,149 @@
...
+        var timeseriesStore = new HashMap();
+        // To check if keyword in query is changed (new query?)
+        var currentKeywords = [""];
+        // To check if geo level in query is changed.
+        var currentGeoLevel = "";
+        // To check if time range in query changed (new query?)
+        var endDate = new Date();
+        // Deducts -1 to create a default value that can't exist
+        endDate.setDate(endDate.getDate() - 1);
+        // Cached time interval, intermediate or complete byTimeRequest result.
```
waans11: I think we don't need to cache the intermediate results.

examples/twittermap/web/public/javascripts/timeseriescache/services.js
ilvry: strict mode is used mainly to make sure variables are declared before use. as in `mapresultcache/services.js`.

waans11: Got it.

examples/twittermap/web/public/javascripts/timeseriescache/services.js
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            return geoIdsNotInCache;
+        };
+
+        /**
+         * Retrieves time-series data from the cache; ignores empty objects
+         */
+        this.getTimeSeriesValues = function (geoIds, geoLevel, timeInterval) {
```
waans11: It seems like it returns all day and count and ignores geoIDs. Why are geoIDs not required? Also, this will return the same day multiple times per GeoID. Is that so?

examples/twittermap/web/public/javascripts/timeseriescache/services.js
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                }
+            }
+            return resultArray;
+        };
+
+        /**
+         * Return time-series histogram data from byTimeRequest result array.
+         */
+        this.getValuesFromResult = function (timeseriesResult) {
```
waans11: Where will this function be used?

ilvry: Used to display the intermediate results. 
intermediate results = cached result + current byTimeRequest result

examples/twittermap/web/public/javascripts/timeseriescache/services.js
ilvry: [line64] var values = timeseriesStore.get(geoIds[i]);
This returns one {day:count} per GeoID, but may return records of the same date for different GeoIDs. (I decided not to merge the count in this case because the histogram library handles it pretty well)

examples/twittermap/web/public/javascripts/timeseriescache/services.js
waans11: OK.

examples/twittermap/web/public/javascripts/timeseriescache/services.js
waans11: OK.

examples/twittermap/web/public/javascripts/timeseriescache/services.js
```diff
@@ -0,0 +1,128 @@
...
+            var geoIdsNotInCache = [];
+            // New query case
+            if (keywords.toString() != currentKeywords.toString() ||
+                timeInterval.start < cachedTimeRange.start ||
+                timeInterval.end > cachedTimeRange.end ||
+                geoLevel != currentGeoLevel) {
+                timeseriesStore.clear();
+                currentKeywords = keywords.slice();
+                currentGeoLevel = geoLevel;
+                cachedTimeRange.start = timeInterval.start;
```
baiqiushi: Let's make sure the assignment is not shallow copy, but should be deep copy. Otherwise, the `cachedTimeRange.start` will always equal to `timeInterval.start`, similar to `end` and `currentKeywords`.

PR #551, opened 2018-07-18
[Issue#515] Frontend refactor - Multilayer management part2(simplfied)

AC-Unicorn:
**1 - Introduce countmap layer service module**
 - `services-multilayer-countmap.js`
 - Similar logic to previous `countmap.js`
 - But maintains the loading polygons features inside itself, e.g. `loadCityJsonByBound()`
 - Using `service` other than previous `controller` mechanism, so that it can be instantiated to multiple instances.

**2 - Introduce multilayer service module**
 - `services-multilayer.js`
 - Factory for different layers which help instantiate a new layer object


examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -226,10 +225,10 @@ angular.module('cloudberry.map')
       $scope.$parent.onEachFeature = onEachFeature;
 
       // Subscribe to moduleManager's events
-      moduleManager.subscribeEvent(moduleManager.EVENT.CHANGE_ZOOM_LEVEL, countMapCommonEventHandler);
-      moduleManager.subscribeEvent(moduleManager.EVENT.CHANGE_REGION_BY_DRAG, countMapCommonEventHandler);
-      moduleManager.subscribeEvent(moduleManager.EVENT.CHANGE_SEARCH_KEYWORD, countMapCommonEventHandler);
-      moduleManager.subscribeEvent(moduleManager.EVENT.CHANGE_TIME_SERIES_RANGE, countMapCommonEventHandler);
+      //moduleManager.subscribeEvent(moduleManager.EVENT.CHANGE_ZOOM_LEVEL, countMapCommonEventHandler);
```
baiqiushi: Since we're not using `countmap.js` anymore, let's not change this file. Later let's just remove this file.

examples/twittermap/web/public/javascripts/map/controllers.js
```diff
@@ -245,7 +245,6 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common','clou
 
     // load geoJson to get city polygons
     $scope.loadCityJsonByBound = function loadCityJsonByBound(onEachFeature, fromEventName, fromEvent){
-
```
baiqiushi: I compared this file with current master, and didn't see any difference between them except the last line `ng-controller`'s value. Let's avoid this kind of submission with files reverted back to original version manually which will result in a lot of format changes highlight, but actually nothing changed. Please use `git checkout -- filePath` to revert a file back to original version.

examples/twittermap/web/public/javascripts/map/services-multilayer-countmap.js
```diff
@@ -0,0 +1,978 @@
...
+              });
+            });
+          }
+          features[id].properties["centerLog"] = (maxLog + minLog) / 2;
+          features[id].properties["centerLat"] = (maxLat + minLat) / 2;
+        }
+      }
+    }
+  
+    function loadCityJsonByBound(onEachFeature,instance){
```
baiqiushi: Per my understanding, now there are two `loadCityJsonByBound` functions defined and called in `services-multilayer-countmap.js` and `controllers.js` respectively. But I wonder why the same logic called twice in two places?

examples/twittermap/web/public/javascripts/map/services-multilayer-countmap.js
```diff
@@ -0,0 +1,984 @@
...
+          instance.layer.addLayer(instance.polygons.stateUpperPolygons);
+          instance.layer.addLayer(instance.polygons.countyPolygons);
+        } else if (instance.status.zoomLevel <= 5) {
+          resetGeoInfo("state");
+          if(!instance.status.init){
+            //TODO publish events
+            //moduleManager.publishEvent(moduleManager.EVENT.CHANGE_ZOOM_LEVEL, {level: instance.map.getZoom(), bounds: instance.map.getBounds()});
+          }
+          if (!instance.status.init) {
+          cloudberry.query(cloudberry.parameters);
```
baiqiushi: This is from old code, not even from current master, let's remove it. I think you need a better development environment.

baiqiushi:
Current know bug is as following: first search something in the `Countmap`, then switch to the `Pinmap`, then move mouse over states, it will highlight polygons.

May need another round of testing before final approval.

chenlica:
@AC-Unicorn @baiqiushi  plan for this PR?

PR #570, opened 2018-08-08
Time-series caching

ilvry:
Merge `timeseries caching branch` to refactored frontend in Master, and resolve conflicts.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
@@ -11,7 +11,7 @@ import edu.uci.ics.cloudberry.zion.model.impl.{DataSetInfo, QueryPlanner}
 import edu.uci.ics.cloudberry.zion.model.schema._
 import edu.uci.ics.cloudberry.zion.model.slicing.Drum
 import org.joda.time.DateTime
-import play.api.libs.json.{JsArray, JsNumber}
+import play.api.libs.json.{JsArray, JsNumber, JsValue, JsObject, Json}
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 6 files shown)


examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -58,60 +58,124 @@ angular.module('cloudberry.map')
...
 
       // Gets the Geo IDs that are not in the map result cache.
       $scope.geoIdsNotInCache = MapResultCache.getGeoIdsNotInCache(cloudberry.parameters.keywords,
         cloudberry.parameters.timeInterval,
         cloudberry.parameters.geoIds, cloudberry.parameters.geoLevel);
 
-      // Batch request with only the geoIds whose map result are not cached yet - partial map result cache hit case
+      // Batch request without time series result - used when the complete time series cache hit, partial map result cache hit case
+      var batchWithoutTimeRequest = {
+        batch: [queryUtil.byGeoRequest(cloudberry.parameters, $scope.geoIdsNotInCache)]
```
baiqiushi: Why this query has no "slicing" check, like batchWithPartialGeoRequest?

ilvry: This query is used with the "complete time series cache hit" case, since the time series results are obtained from the cache immediately (no slice), we can get the corresponding final map result.

ilvry: Oh, I see. In this case, should the time-series histogram displayed in slices? (to be synced with this request)

baiqiushi: Yes. I think that's what we decided.

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ProgressiveSolver.scala
```diff
@@ -98,7 +98,24 @@ class ProgressiveSolver(val dataManager: ActorRef,
...
-        reporter ! Reporter.PartialResult(curInterval.getStartMillis, boundary.getEndMillis, 1.0, queryGroup.postTransform.transform(JsArray(returnedResult)))
+
+        val timeInterval = Json.obj(
+          "timeInterval" -> Json.obj(
+            "start" -> JsNumber(curInterval.getStart().getMillis()),
+            "end" -> JsNumber(boundary.getEnd().getMillis())
+        ))
+        // for query with slicing request, add current timeInterval information in its query results.
+        val infoValue = queryGroup.postTransform.transform(JsArray(returnedResult))
+        var results : JsValue = JsObject(Seq.empty)
```
baiqiushi: Is it possible if we combine this line with the following line like this:
```
val results : JsValue = infoValue match {
...
}
```

PR #576, opened 2018-08-16
added README.md

GaeulG:
I added annotation for korean twittermap.


baiqiushi:
This is only for the purpose to get familiar with the process of creating a PR.

Issue #585, opened 2018-08-22
Add an api for Cloudberry to check the existence of view


lengyyy:
## Overview
- For **Coundberry Server**: it aims to add a new **api** so that Twittermap server can check whether a query can solve by view (cached result) or not.
- For **Twittermap application**: it aims to **disable** the button of sidebar when view is not established. And enable sidebar when query can solved by the view. It can improve the user experience about content in the sidebar.
#### For cloudberry
-  **RequestRouter.scala**
 Make the name of client and other variable names more general.
- **Cloudberry.scala** 
Add new web socket for checking whether a query can be solved by view.
- **routes**
Add address for new api in routes
- **JsonParser.scala**
Add a function to get queryID
- **QueryPlanner.scala**
Add a function to return whether there is matched views for a query.
 Extract a function to find the matched views for a query.
- **ViewStatusClientTest.scala**
Three tests for ViewSatusClient.
- **ViewStatusClient.scala**
**1.** This is a reactive client which checks whether a query can be solved by the existed view.
**2.** When client receiving message: parse the json to queries, and for each query let planner to decide  whether it can be solved by view. 
**3.** Return the "isQuerySolvableByView" result with queryID to frontend

 #### For twittermap
-  **TwitterMapApplication.scala**
Add a WebSocket that send query to Cloudberry
- **application.conf**, **common/services.js**, **routes**
Add related address for new api
- **index.scala.html** 
Add id for sidebar button.
- **sidebar/controllers.js**
**1.** Add a WebSocket that send query to Cloudberry
**2.** When the keywords changed, we need to: clear previous timer, close and disable sidebar, and set a new timer for new keywords. The timer sends query to check whether it is solvable, every one second.
**3.** When receiving messages from websocket, check its queryID and result. If queryID is matched and result is true, enable the sidebar button and clear timer.

## Result
In this demo: 
- "code" is a keyword without cached view
- "water" is a keyword already cached.

![imageedit_1_7641184993](https://user-images.githubusercontent.com/26839941/44494410-72346200-a620-11e8-91b8-e0b309b15cb9.gif)

## Resources
To add an api to Cloudberry, and let Twittermap communicate with it, some related document of the [Play Framework](https://www.playframework.com/documentation/2.5.x/ScalaWebSockets), and [akka Framework](https://doc.akka.io/docs/akka/2.5/guide/tutorial.html) may be help.

## To Do List
[...]

PR #588, opened 2018-08-23
[issue #585] Add an api for Cloudberry to check the existence of view

lengyyy:
## Overview
Please refer to issue #585.
- For **Coundberry Server**: it is to add a new **api** so that Twittermap server can check whether a query can solve by view (cached result) or not.
- For **Twittermap application**: it is to **disable** the button of sidebar when view is not established. And enable sidebar when query can solved by the view, which can improve the user experience about content in the sidebar.

#### For cloudberry
-  **RequestRouter.scala**
 Make the name of client and other variable names more general.
- **Cloudberry.scala** 
Add new web socket for checking whether a query can be solved by view.
- **routes**
Add address for new api in routes
- **JsonParser.scala**
Add a function to get queryID
- **QueryPlanner.scala**
Add a function to return whether there is matched views for a query.
 Extract a function to find the matched views for a query.
- **ViewStatusClientTest.scala**
Three tests for ViewSatusClient.
- **ViewStatusClient.scala**
**1.** This is a reactive client which checks whether a query can be solved by the existed view.
**2.** When client receiving message: parse the json to queries, and for each query let planner to decide  whether it can be solved by view. 
**3.** Return the "isQuerySolvableByView" result with queryID to frontend

 #### For twittermap
-  **TwitterMapApplication.scala**
Add a WebSocket that send query to Cloudberry
- **application.conf**, **common/services.js**, **routes**
Add related address for new api
- **index.scala.html** 
Add id for sidebar button.
- **sidebar/controllers.js**
**1.** Add a WebSocket that send query to Cloudberry
**2.** When the keywords changed, we need to: clear previous timer, close and disable sidebar, and set a new timer for new keywords. The timer sends query to check whether it is solvable, every one second.
**3.** When receiving messages from websocket, check its queryID and result. If queryID is matched and result is true, enable the sidebar button and clear timer.

## Result
In this demo: 
- "code" is a keyword without cached view
- "water" is a keyword already cached.

imageedit_1_7641184993


baiqiushi:
Firstly, for Cloudberry layer code, we need test cases for corresponding new logic, and for this case let's add at least one test case for `ViewStatusClient` which should be similar to `SimpleBerryClientTest`. @lengyyy 

Then let's go to the details to review this whole PR.

lengyyy:
`ViewStatusClientTest` is added, and the travis-ci passes.

chenlica:
@lengyyy Please add the comments we provided by @waans11 and me today in the group meeting.

lengyyy:
The @chenlica and @waans11  comments on Sept 6th, group meeting:

- [ ] Add comment for the code.
- [ ] Change function name and variable name.
- [ ] Avoid duplicate checking of existence of tables in `ViewStatusClient.scala`
- [ ] Add query id in the api format, to avoid asynchronism problem.
- [ ] Add some logic description of the codebase in issue

lengyyy:
@waans11 comments on Sept 7th:

- [ ]  Add more comments to explain the reason for necessary of code
- [ ]  Add a test case for the "isQuerySolvableByView: false" situation
- [ ]  QueryID should be String instead of Int
- [ ]  Extract a function for a duplicate using code
- [ ]  Change the name of "timer" and others to a more specific one for sidebar controller

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/ViewStatusClient.scala
```diff
@@ -0,0 +1,108 @@
...
+      handleRequest(json)
+    case (json: JsValue, transform: IPostTransform) =>
+      handleRequest(json)
+  }
+
+  // Handle the Request in following step
+  private def handleRequest(json: JsValue): Unit = {
+    // 1. Use jsonParser to parse the json to get datasets name and queryID
+    val datasets = jsonParser.getDatasets(json).toSeq
+    val queryID = jsonParser.getQueryID(json)
```
baiqiushi: About queryID part, Cloudberry query format has already has an ID parameter in `transform` object, (refer to https://github.com/ISG-ICS/cloudberry/pull/471/files). So we can reuse that as the queryID functionality.

examples/twittermap/web/public/javascripts/sidebar/controllers.js
```diff
@@ -11,6 +11,58 @@ angular.module("cloudberry.sidebar", ["cloudberry.common"])
...
+    // Timer for sending query to check whether it can be solved by view
+    $scope.timerCheckQuerySolvableByView = null;
+
+    // queryID used to identify a query, which is sent by timer
+    $scope.nowQueryID = null;
+
+    // A WebSocket that send query to Cloudberry, to check whether it is solvable by view
+    var wsCheckQuerySolvableByView = new WebSocket(cloudberryConfig.checkQuerySolvableByView);
+
+    function closeRightMenu() {
```
baiqiushi: There's a small bug here: this `closeRightMenu` function needs to include the operations of setting `isHashTagOpen` and `isSampleTweetsOpen` flags to be `false` as well. Otherwise, although you close the sidebar, the `hashtag` or `sample tweets` query will still be sent to Cloudberry.

lengyyy: For open/ close sidebar, I find a bug in current code.
image

In the code, we use different button for open and close sidebar separately. 
- When click `hambugerButton`, it opens sidebar, and changes the flags by `showOrHideSidebar` function.
- When click `closebtn`, it only closes sidebar, but **not** change the flags.

I think it is because we use only one button for both open and close in previous version. 
Now we use two button, we need to change the code.

lengyyy: Should I fix the bug in this PR?

baiqiushi: Oh, I see. It was introduced by the `hamburger` button. Let's fix it since it's just one line fix. I think just adding the `ng-click="showOrHideSidebar(-1)"` will solve the problem. Please correct me if I'm wrong.

baiqiushi: And we don't need `click = click * -1` anymore. Since open and close buttons are separated.

lengyyy: I don't think we need variable `click` anymore. So I remove the code about click. 
Also, I move the js code from `index.scala.html` to `sidebar/controller.js` to avoid duplicate function.

lengyyy:
I don't think we need variable `click` anymore. So I remove the code about click.
Also, I move the js code for open/close sidebar from `index.scala.html` to `sidebar/controller.js` to avoid duplicate function.

baiqiushi:
Well done! Thank you!

Issue #575, opened 2018-08-15
Replace byGeoRequest with byGeoTimeRequest


ilvry:
## Overview
The new byTimeRequest (will be named as **byGeoTimeRequest**) returns `{geoID, day, count}`. Caching datasets `dsStatePopulation`, `dsCountyPopulation` and `dsCityPopulation`: `{geoID, population, …}`, we can obtain `map results` (**byGeoRequest** results) `{geoID, count, population}` from byGeoTimeRequest.

## HTTP Request
There are currently 2 approaches to this issue:
(a) Send JSON request to datasets
**(b) HTTP request to local files**

Although the `retrieve from database design` would work, we decided on the taking the http request approach for the following reasons:
- **(1)** Low ratio of access to the datasets
Since the main goal is to replace the `byGeoRequest` with `byGeoTimeRequest` and a `population cache module`, the population dataset is not accessed as frequently as before. The `byGeoRequest` requests from population datasets once per user movement, whereas in the current case, each geoId is accessed at most once during one section of using Cloudberry.
- **(2)** Faster local access
Http request to local population data is in most cases faster than a json request, it also does not require the time and space for data injection to the database.

Therefore, it may not be necessary  to inject them to the database.
p.s. A similar case is the `geo polygon load`.

## Experimental Results
Based on the results in the following link, this change will decrease the time on waiting for cloudberry server response:
https://docs.google.com/spreadsheets/d/1fmmRAUjLcdHqb7C6Q8BGLuGU_PM5rzFLiYRgQBNKtoA/edit?usp=sharing
________

- [x] Experiment: byTimeRequest+byGeoRequest batch VS. GeoTimeRequest
- [x] Implement approach (b)
- [x] Make sure the opulation data is lazy and dynamic loaded
- [x] Add to Cloudberry

baiqiushi:
#589

PR #589, opened 2018-08-27
[issue#575] Population dynamic lazy load

ilvry:
## Overview
Replace original `byGeoRequest` with:
- (1) `byGeoTimeRequest` and,
- (2) obtain `{geoID, population}` through http request.
Doing so will eliminate registering population datasets and byGeoRequest response time.

Please refers to #575 for more details and experimental result.
_________
The implementation consists of three major changes:
- (1) **front-end**: Replace `byGeoRequest` result from `byGeotimeRequest` result and `population cache result.`
- (2) **front-end**: New `population cache module`
- (3) **twittermap web server**:  
Prefetch `state` and `county` level population data; Load `city` level population data as needed.
The `city` population load follows two rules:
(a) is loaded as **lazy** as possible, similar to the sidebar query case
(b) is loaded **dynamically** based on current visible geo locations, similar to the geo polygon cache case
## Testing
Is preferred to test with `us.city.path = "/public/data/city.json”` by commenting line108 in `application.conf` and uncomment line110.


cloudberry/neo/conf/production.conf
```diff
@@ -79,5 +79,6 @@ bounded-mailbox {
 asterixdb.url = "http://actinium.ics.uci.edu:19002/query/service"
 asterixdb.lang = SQLPP
 us.city.path="/public/data/city.json"
+us.citypop.path="/public/data/allCityPopulation.json"
```
baiqiushi: Is this used in configuration file of `neo`?

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -197,6 +209,49 @@ object TwitterMapApplication {
...
+  }
+
+  /**
+    * Find the population data data for each geoId in from the sorted geoIds list.
+    *
+    * @param sortedCityIds Sorted list of cities in current boundary
+    * @param citiesPopulation List of all cities' population data
+    * @return List of cities in sortedCityIds' population data
+    */
+  def sortedListBinarySearch(citiesPopulation: List[JsValue], startIndex: Int, endIndex: Int,
```
baiqiushi: So we need the `allCityPopulation.json` sorted by cityID in ascending order, correct?

examples/twittermap/web/public/javascripts/map/controllers.js
```diff
@@ -318,7 +317,30 @@ angular.module('cloudberry.map', ['leaflet-directive', 'cloudberry.common','clou
             });
         }
     };
-    
+
+    // load population Json to get state and county polygons, store in countmap cache.
+    $scope.loadPopJsonFiles = function loadPopJsonFiles(onEachFeature) {
```
baiqiushi: `onEachFeature` parameter is not necessary for this function, it was used for creating polygon objects by Leaflet.js library to register events listeners for `loadCityPolygons` function and etc.

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
ilvry: Yes.

examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -183,17 +124,33 @@ angular.module('cloudberry.map')
             resultSet[0] = [];
             cloudberry.commonTimeSeriesResult = TimeSeriesCache.getValuesFromResult(cloudberry.timeSeriesQueryResult).concat(
               TimeSeriesCache.getTimeSeriesValues(cloudberry.parameters.geoIds, cloudberry.parameters.geoLevel, requestTimeRange));
-            cloudberry.countmapMapResult = resultSet[1].concat(cloudberry.countmapPartialMapResult);
+            if (cloudberry.parameters.geoLevel === "city") {
+              var newCities = [];
+              for (i = 0; i < cloudberry.parameters.geoIds.length; i++) {
+                newCities.push(cloudberry.parameters.geoIds[i]);
+              }
+              PopulationCache.loadCityPopulationToCache(newCities).done(function(){
```
baiqiushi: There may be a problem here. For this cache not all hit case, the callback function of `cloudberryClient.send` could be called multiple times if `slice` option is `ON`, then this `loadCityPopulationToCache` function could be called multiple times also, which might not be necessary.

examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -183,17 +124,33 @@ angular.module('cloudberry.map')
...
+              TimeSeriesCache.arrayToStore($scope.geoIdsNotInCache, cloudberry.timeSeriesQueryResult, cloudberry.parameters.geoLevel)).concat(
+              cloudberry.countmapPartialMapResult);
+            }
           }
           // When the query is executed completely, we update the map result cache and time series cache.
           if((cloudberryConfig.querySliceMills > 0 && !angular.isArray(resultSet) &&
             resultSet['key'] === "done") || cloudberryConfig.querySliceMills <= 0) {
-            MapResultCache.putValues($scope.geoIdsNotInCache, cloudberry.parameters.geoLevel,
-              cloudberry.countmapMapResult);
+            MapResultCache.putValues($scope.geoIdsNotInCache, cloudberry.parameters.geoLevel, cloudberry.countmapMapResult);
```
baiqiushi: Similar here, `MapResultCache` may be useless.

examples/twittermap/web/public/data/allStatePopulation.json
```diff
@@ -0,0 +1,261 @@
+[
```
baiqiushi: To reduce the network overhead, I think for these 3 json files, those attributes: `name` in `allStatePopulation.json`; `name`. `stateID`, `stateName` in `allCountyPopulation.json`; and `name`, `countyID`, `countyName`, `stateID`, `stateName` are unnecessary, since all the IDs are unique themselves. And if we get rid of them, we may also save much `TwittermapAppalication` memory (since you load them into memory on the server side).

examples/twittermap/web/public/javascripts/map/countmap.js
```diff
@@ -69,108 +69,49 @@ angular.module('cloudberry.map')
...
+
+        var emptyStore = new HashMap();
+        if (cloudberry.parameters.geoLevel === "city") {
+          var newCities = [];
+          for (i = 0; i < cloudberry.parameters.geoIds.length; i++) {
+                newCities.push(cloudberry.parameters.geoIds[i]);
+          }
+          PopulationCache.loadCityPopulationToCache(newCities).done(function(){
+            cloudberry.countmapPartialMapResult = MapResultCache.getValues(cloudberry.parameters.geoIds, cloudberry.parameters.geoLevel);
+            cloudberry.countmapMapResult = PopulationCache.getCountMapValues($scope.geoIdsNotInCache, cloudberry.parameters.geoLevel,
```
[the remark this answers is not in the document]
baiqiushi: For this reason and according to my understanding of this new design, my suggestion is that: All `countmap` results will come from `TimeSeries` module, all `TimeSeries` module queries will be handled by `TimeSeries/controllers.js-TimeSeriesCtrl` and once it gets some update on `TimeSeries` results, it notifies the `Countmap.js-countMapCtrl` to update its results. Once `Countmap.js-countMapCtrl` gets message saying results should be updated, it will call `TimeSeriesCache` module's API function to get the `{state/county/city, count}` part data, and call `PopulationCache` module's API function to add `population` attribute to the data and accumulate to the final `{state/county/city, count, population}` data format.

Then I think the logic of this part will look more clear and easy to read.

chenlica:
@baiqiushi  plan for this PR?

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -290,6 +302,49 @@ object TwitterMapApplication {
...
+        sortedListBinarySearch(citiesPopulation, startIndex, thisIndex, sortedCityIds, sortedCityIndex, resultArr)
+      } else if (thisCityId < sortedCityIds(sortedCityIndex)) {
+        sortedListBinarySearch(citiesPopulation, thisIndex + 1, endIndex, sortedCityIds, sortedCityIndex, resultArr)
+      } else { // Current cityId sortedCityIds(sortedCityIndex) is find in citiesPopulation
+        val currPop : JsValue = JsObject(Seq("cityID" -> JsNumber(thisCityId),
+                                             "population" -> JsNumber((citiesPopulation(thisIndex) \ "population").as[Int])
+                                            ))
+        if (sortedCityIndex < sortedCityIds.length - 1) {
+          sortedListBinarySearch(citiesPopulation, thisIndex + 1, 29833, sortedCityIds, sortedCityIndex + 1, resultArr :+ currPop)
+        } else {
```
baiqiushi:
```suggestion
        } else {
```
This is wrong, only the population data of the **last** city in the `sortedCityIds` can be appended to the `resultArr`.

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -290,6 +302,49 @@ object TwitterMapApplication {
...
+          resultArr :+ currPop
```
baiqiushi: We need to put this append operation just after we get `currPop` at line 339.

baiqiushi: Sorry, I missed that your do the append operation also in the recursive call of the function at line 340. Then it should work.

PR #590, opened 2018-08-30
Fix the bug about assign BooleanField

lengyyy:
The bug is in `cloudberry/zion/model/schema/AbstractSchema.scala` line 78.

image

It should be **"case DataType.Boolean => BooleanField(name, field.isOptional)"**


baiqiushi:
Let's test this simple fix PR on backup server once just for safety. Then we can merge it.

PR #595, opened 2018-09-19
Fix the bug of counting the total number of tweets in the dataset

dayuebai:
### Example
* Dataset: 47,000 tweets (from 01/01/2016 to 05/05/2017)

### Bug
* When cloudberry first starts, cloudberry sends a request to count the number of tweets in the dataset. Cloudberry will get the response: ```[{"count": 47000}]```. Then, for some reason, cloudberry is out of connection. After user refreshes the web page, cloudberry sends another query to count the number of tweets created from 05/05/2017 (inclusive) to current time (exclusive). After getting the result (E.g. ```[{"count": 2}]```), cloudberry adds the two number together and upserts ```47002``` into berry.meta. In fact, in the second query, cloudberry should not count the number of tweets created on 05/05/2017, because this number is included in the first response: ```[{"count": 47000}]```.

### Solution
* In the second query, Cloudberry will only count the number of tweets created from 05/05/2017 (exclusive) to current time (inclusive). In this case, whenever Cloudberry collects the statistics afterwards, the result should be correct.

### Code implementation: 
* Delete the existing wrong ```FilterStatement```
* Add a ```FilterStatement``` with ```Relation.>``` to exclude the start date.
* Add a ```FilterStatement``` with ```Relation.<=``` to include the end date.
* Add the two new ```FilterStatement``` into the query.

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
@@ -69,9 +69,10 @@ class OriginalDataAgent(val dataSetInfo: DataSetInfo,
     }
     val temporalSchema = dataSetInfo.schema.asInstanceOf[Schema]
     val now = DateTime.now().minusMillis(1)
-    val filter = FilterStatement(temporalSchema.timeField, None, Relation.inRange, Seq(start, now).map(TimeField.TimeFormat.print))
+    val startFilter = FilterStatement(temporalSchema.timeField, None, Relation.>, Seq(start).map(TimeField.TimeFormat.print))
+    val endFilter = FilterStatement(temporalSchema.timeField, None, Relation.<=, Seq(now).map(TimeField.TimeFormat.print))
     val aggr = GlobalAggregateStatement(AggregateStatement(temporalSchema.fieldMap("*"), Count, Field.as(Count(temporalSchema.fieldMap("*")), "count")))
-    val queryCardinality = Query(dbName, filter = Seq(filter), globalAggr = Some(aggr))
+    val queryCardinality = Query(dbName, filter = Seq(startFilter, endFilter), globalAggr = Some(aggr))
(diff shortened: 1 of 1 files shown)


ChankiWu:
LGTM. Please do more tests to make sure the logic works fine !

chenlica:
@dayuebai @baiqiushi  plan for this PR?

dayuebai:
> @dayuebai @baiqiushi plan for this PR?

@ChankiWu has reviewed this PR. Waiting for merge.

chenlica:
@baiqiushi @waans11 should we merge it?

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/OriginalDataAgent.scala
```diff
@@ -69,9 +69,10 @@ class OriginalDataAgent(val dataSetInfo: DataSetInfo,
     }
     val temporalSchema = dataSetInfo.schema.asInstanceOf[Schema]
     val now = DateTime.now().minusMillis(1)
-    val filter = FilterStatement(temporalSchema.timeField, None, Relation.inRange, Seq(start, now).map(TimeField.TimeFormat.print))
+    val startFilter = FilterStatement(temporalSchema.timeField, None, Relation.>, Seq(start).map(TimeField.TimeFormat.print))
```
baiqiushi: Sorry for the delay. Here we think that the original usage of `inRange` logical operator is good practice to maintain the same semantic meaning for different DB adapters no matter how the DB adapters  implement this operator.

baiqiushi: And after discussion with @dayuebai, we agreed that the problem is caused by the first cardinality counting query was using the >= and <= operators but later collecting states queries are using this `inRange` operator which is usually implement as >= and <. So that the intersect one day's records are counted twice. So we proposed the solution to see if we can modify the first query to also use this `inRange` operator.

chenlica: It's good to maintain the generality of the function.

PR #600, opened 2018-09-28
Korean twittermap

LeeEunah:
1. Change data files. (state.json, county.json, city.json)
- Because of big size files, simplified them to 25%.
2. Move coordinates to show Korea. (lat="35.95" lng="127.76" zoom="7")
- Change the zoom value of reset button.
- Change zoomLevel.
3. Solve geotag error
- The date between Korean datas and AsterixDB doesn't match. I set them to read as Locale.US.
- Some data has null value in coordinate. So I handled NullPointError by using the coordinates in bounding_box.

4. Demo
ezgif com-video-to-gif


chenlica:
@waans11 @AC-Unicorn @baiqiushi  plan for this PR?

PR #607, opened 2018-10-31
HIVMap

baiqiushi:
A customized version of Twittermap by sending `hiv` as the keyword by default once the Twittermap page is loaded into browser.


chenlica:
Please go ahead to do the merge!

PR #612, opened 2018-11-07
[Bug fix] Live tweets

AC-Unicorn:
- Twitter API used to name rendered tweet object as "twitterwidget" now it changed to "twitter-widget", this change will cause bug in live tweets.


examples/twittermap/web/public/javascripts/sidebar/controllers.js
```diff
@@ -244,10 +276,14 @@ angular.module("cloudberry.sidebar", ["cloudberry.common"])
...
-        cloudberry.parameters.isSampleTweetsOpen = false;
-        cloudberry.parameters.isHashTagOpen = false;
+        $scope.isSampleTweetsOpen = false;
+        $scope.isHashTagOpen = false;
+        window.clearInterval(sendQueryLoop);
+        window.clearInterval($scope.liveTweetsLoop);
+        clearTimeout(secondLiveTweetQueryTimeOut);
       }
       else {
+        resumeLiveTweet();
```
baiqiushi: Firstly, this `resumeLiveTweet()` call should be inside `showTab` function, because here you don't know whether the `SampleTweet` tab is open or not.
Secondly, even inside `showTab`, it may cause some issue because every time `showTab` is called, the `handleSidebarQuery()` is called. So, we need to consider whether the same logic will be called twice.

PR #618, opened 2018-11-15
[Frontend] Live tweets improvement 

AC-Unicorn:
To enhance user experience, I did following adjustment on live tweets,

1. Instead send query every 3 seconds to get latest 1 tweet, now we send query every 60 seconds to get 10 tweets. 

2. Made the send query and update rate configurable by adding parameter in app/controllers/TwitterMapApplication.scala

3.Clean live tweet update and send query loop every time user switch tab or close sidebar, and resume the loop when user switch back.


chenlica:
Since the live tweet feature is so valuable, I wonder whether we can show it every time a user does a search, even before the user clicks the "hamburger" button?

chenlica:
Yes, that's what I meant.

baiqiushi:
We need to talk about this part F2F.

AC-Unicorn:
> We need to talk about this part F2F.

Let's meet on monday

PR #559, opened 2018-07-25
[issue#557] Cloudberry carry time-interval with each sliced query result

ilvry:
Cloudberry carry the slicing `timeInterval` information along with the slicing results.

An example query result for query with `cloudberryConfig.querySliceMills > 0`:
> {
category: "batchWithPartialGeoRequest"
id: "batchWithPartialGeoRequest"
timeInterval: {start: 1493366400000, end: 1493539200000}
value: [Array(0), Array(4), Array(0)]
> }

Otherwise, the result will not include the  `timeInterval` variable:
> {
category: "sample"
id: "sample"
value: [Array(0)]
> }

PR #623, opened 2018-11-25
Enable A New Flag `returnDelta` Inside `option` Parameter Of A Query

baiqiushi:
## Overview
There's an `option` parameter in a Cloudberry query, which is used for query slicing parameters, such as `slicemillis` (a.k.a pace milliseconds in Drum). However, our current way to return sliced results every time is to return all accumulated results we've got from the DB until now, i.e. ∑(r1+r2+r3+...+rN). Now to support some non-aggregation queries, which has accumulated results size growing linearly, we need to set up a new flag parameter called `returnDelta` inside `option`that if it's `true`, Cloudberry will return only the delta result of sliced queries. If not specified, the default value of `returnDelta` is `false`.

A query may look like the following:
```
{
    dataset: ...,
    filter: { ... },
    select: { ... },
    option: {
        slicemillis: 2000,
        returnDelta: true
    }
}
```

Updated 2018-12-02
## Known Issues:
- [x] ***Important** 
    - **Symptom**: Total number of all `delta` records returned from Cloudberry does NOT equal to one time query result without `slicing`.
    - **Theory**: In current implementation of `Drum`, due to its `return all results until now` semantic, whenever Cloudberry got a new partial result from DB, it will drop all results queued inside Cloudberry which have not been sent back to the client. However, in our new `return delta result this time` semantic, some intermediate results could be dropped.

- [ ] **Low Priority**: To support time series bar cache feature, we added a `timeInterval` field into the sliced results, #559. To make this `timeInterval` consistent semantically, we need to change the `start` value for each `delta` result as well.


chenlica:
@AC-Unicorn @baiqiushi is it almost done?

Issue #627, opened 2018-11-30
Convert the timebar with a new library "highcharts"


dongchime:
Now the timebar uses dc js library.
we want to change the chart to stacked area and there is a bug that when sometimes user types a keword, strange line is shown.
So I decide to change the timebar with better library. 

https://www.highcharts.com/

chenlica:
Will close it since presumably @AC-Unicorn has a difference solution.

PR #628, opened 2018-11-30
[issue #627] Convert timebar library

dongchime:
- [x] Render the timebar
- [x] When the map changes, the timebar also changes.
- [ ] When timebar's range is zoomed, the map changes.


baiqiushi:
According to our further investigation, high-chart.js is not suitable for our demand.

PR #636, opened 2019-01-10
[Fix Bug] handleAvg exception when returned result from DB is empty

baiqiushi:
* Add a check, if the returned result from DB is empty, return immediately
* Minor optimization of following if-else code

### DIFF
--- a/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryPlanner.scala
+++ b/cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryPlanner.scala
@@ -288,6 +288,11 @@ object QueryPlanner {
   // Handle avg function
   def handleAvg(result:JsArray):JsValue =
   {
+    // If result is empty, return immediately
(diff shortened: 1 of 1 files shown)


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/QueryPlanner.scala
```diff
@@ -302,25 +307,24 @@ object QueryPlanner {
 
     // No avg fields need to be merged, return immediately
     if (mergedField.isEmpty) {
-      result
+      return result
     }
-    else {
-      //merge sum and count to avg
-      val handledRows = result.value.map(row => {
```
baiqiushi: I didn't change any of the following code, just get rid of the `else` wrap.
* Previous:
```
if (something empty) {
    result
}
else {
    do something
    result
}
```
* Now:
```
if (something empty) {
    return result
}

do something
result (This is syntax of scala, the last sentence is the returned result by default, no explicit return statement needed)
```

Issue #647, opened 2019-02-05
[Backend] Parallelize geotagging


baiqiushi:
# Overview
Previously, while geotag module supported multi-threads pool parallelization, it had some issue when they write the output adm string to the same console. To solve this issue, we made an initial attempt to synchronize the output stream and initial experiments showed correctness and speedup.

# GeoTagging pipeline 
![image](https://user-images.githubusercontent.com/10699892/53544994-d4dd4d80-3adc-11e9-806d-c3906b6909be.png)

# Plan
 - [x] Create a PR for this simple patch
 - [x] Test it on the 5-node cluster running old production database
 - [x] Merge it

baiqiushi:
#671

PR #648, opened 2019-02-05
[Issue #647] Synchronize the write out operation of multi-threaded tagOneTweet

baiqiushi:
#647


baiqiushi:
This solution is proved to be buggy by experiment.

PR #652, opened 2019-02-11
Revert "Fix date format timezone bug in AbstractSchema"

baiqiushi:
Reverts ISG-ICS/cloudberry#632

# Overview
Since current ingestion script does not add timezone info to the `adm` instance pushed to AsterixDB, the tweets in current production DB is all hardcoded by `UTC` timezone, which should be `PST` timezone instead. And this problem causes current DB will not work properly with this fix merged.
Therefore, this PR could be merged once the problem stated above is solved.


baiqiushi:
We decided to fix the ingestion bug. So no need to revert this PR.

PR #653, opened 2019-02-13
[FrontEnd]Live tweets remove duplicate 

AC-Unicorn:
A small modification to remove duplicate tweets in sidebar.

### DIFF
--- a/cloudberry/project/dependencies.scala
+++ b/cloudberry/project/dependencies.scala
@@ -26,7 +26,7 @@ object Dependencies {
   val zionDependencies: Seq[ModuleID] = Seq(
     "joda-time" % "joda-time" % "2.9.3",
     "com.typesafe.akka" %% "akka-actor" % "2.4.4",
-    "com.github.noraui" % "ojdbc8" % "12.2.0.1",
+    "org.zenframework.z8.dependencies.commons" % "ojdbc6" % "2.0",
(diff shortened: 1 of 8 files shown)


chenlica:
@AC-Unicorn For the change of OJDBC, please raise a separate PR.

baiqiushi:
@AC-Unicorn One question, why don't we eliminate the duplicates on the backend? So that we don't waste the network effort to deliver the duplicates to the frontend.

AC-Unicorn:
> @AC-Unicorn One question, why don't we eliminate the duplicates on the backend? So that we don't waste the network effort to deliver the duplicates to the frontend.

So here are my concernings.
1. We should clean the duplicate set periodically, otherwise, the length of the set will increase quickly. And I think the best timing is when there's a new event triggered by the frontend because user actioned then some parameters of the query will be changed ( like drag event, geoinformation will change)

Note: If we never clean the duplicate set, when we search for rare keywords, we'll get no result when the second time we search it.

2. It's easier to detect the parameter change in the frontend because we have the event handler. But for the backend, we need to extract information from queries and cache the last query ( to see if the parameter changed or the query is just a repeated query after 60 seconds).

baiqiushi:
OK. I see the point. Then let's use frontend logic for now.

PR #655, opened 2019-02-14
[TM] Minor enhancement to streamFeed.sh

baiqiushi:
Add US location parameters to streamFeed.sh and make default ingestion to local AsterixDB

### DIFF
--- a/examples/twittermap/script/streamFeed.sh
+++ b/examples/twittermap/script/streamFeed.sh
@@ -8,6 +8,5 @@ Your Consumer Secret
 Your Access token
 -ts
 Your Acces Secret Token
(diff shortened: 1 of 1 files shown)


examples/twittermap/script/streamFeed.sh
```diff
@@ -8,6 +8,5 @@ Your Consumer Secret
 Your Access token
 -ts
 Your Acces Secret Token
--fo
--tr trump \
+-loc -173.847656,17.644022,-65.390625,70.377854 \
```
chenlica: Add a comment to the code to explain the purpose of these numbers?

PR #657, opened 2019-02-15
[Frontend] Set Live tweet feature open as default

AC-Unicorn:
## Keep the sidebar open unless user clicked the close button
We used to close sidebar after user issued a search query because we need to check if the query is solvable by view table and disable hamburger button for a while. The current design is sidebar will open after keyword query been issued.


chenlica:
Please show a demo in our meeting on Friday before @baiqiushi does a review.

PR #659, opened 2019-02-15
[FrontEnd] Issue query by default if url contains keyword parameter

baiqiushi:
Using url parameter keyword as the default search keyword and issue query once page loaded.

screen shot 2019-02-15 at 11 51 13 am


chenlica:
Add a screenshot with the effect of adding a keyword?

PR #660, opened 2019-02-16
[Frontend]Disable hashtag

AC-Unicorn:
##  Elimilate disable Hamburger button function
First I removed the disable hamburger button function, we used to disable the hamburger button for checking if the query is solvable by the view table. Although this function is only used for showing hashtags,  I retained this feature, instead of disabling the hamburger button, I disabled hashtag button when checking the view. When Hashtag button is disabled, the button turns to grey and unclickable. 
image
After we recieved the query result, the hashtag button will restore to blue.


chenlica:
@baiqiushi can review it.

PR #661, opened 2019-02-17
[Frontend]move map choice to left top corner

AC-Unicorn:
Move the map choice box from right bottom corner to left top corner. 
image


chenlica:
Can we add a screenshot?

PR #662, opened 2019-02-18
[Minor Fix] Move comments in streamFeed.sh

baiqiushi:
Comments in the arguments string will block later arguments input into the program.


chenlica:
I believe the comments are valuable.  Instead of removing this useful info, can we figure out the right syntax to include it?  That's a better solution!

chenlica:
LGTM now!

PR #669, opened 2019-02-28
[FrontEnd] Add logic to reconnect websocket if error occurs

baiqiushi:
# Overview
When network is not stable, `ws = new Websocket("ws://x.x.x.x")` will throw an error. This PR is to solve this problem by keep reconnecting once error occurs.


chenlica:
Can we ask @ruig2 to review it?

ruig2:
Talked to @baiqiushi about this one. It looks good to me.

Note that Github shows some lines are added but in fact not.

PR #671, opened 2019-03-04
Increase buffer size to improve geoTagger's performance

AC-Unicorn:
# Background 
Our current implementation of parallel geotagging is not efficient. It geotags a batch of tweets in parallel and waits for all these results to be completed before processing next batch. Therefore, a lot of time is wasted for the sync operation. To reduce the number of sync operations, we can increase the batch size as specified by the buffer size. 

Experiments show that this approach has a significant effect on the processing speed.  The results are as follows:

Figure 1, the speed comparison with different buffer sizes (including only the geoTagging step without ingesting into the database)
image

[Experiment data sheet(public)](https://docs.google.com/spreadsheets/d/1FTFSo-bD4vMtIFzLVZxvQwsGtZvJYEQH86fEvx-E040/edit?usp=sharing)


examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/TwitterJSONTagToADM.scala
```diff
@@ -63,11 +64,11 @@ object TwitterJSONTagToADM {
     val thpool = Executors.newFixedThreadPool(threadNumber)
     implicit val ec: ExecutionContext = ExecutionContext.fromExecutor(thpool)
 
-    val buffer = new ArrayBuffer[String](threadNumber)
+    val buffer = new ArrayBuffer[String](bufferSize)
```
chenlica: Add comments to the codebase: "We use a buffer to store a batch of records for multiple threads to process."

chenlica:
After @baiqiushi reviews and approves it, please go ahead to do the merge.

PR #674, opened 2019-03-15
[Frontend]Auto complete scala

AC-Unicorn:
### Background ###
Our current auto-complete feature now relies on a python proxy server, which is annoying, because everytime we apply change on production server will need to restart the proxy. Now I change it to a scala function.


baiqiushi:
@AC-Unicorn Let's delete the previous proxy scripts as well.

PR #678, opened 2019-04-14
[Frontend]Live tweet improvement

AC-Unicorn:
- retain twitter search API as a sample tweets source.

- Add a switch in `application.conf` for the user to choose sample tweets source.

### DIFF
--- a/examples/twittermap/web/app/controllers/TwitterMapApplication.scala
+++ b/examples/twittermap/web/app/controllers/TwitterMapApplication.scala
@@ -22,6 +22,7 @@ import java.net.{HttpURLConnection, URL, URLEncoder}
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.io.Source
+import scala.util.matching.Regex
 
 
 @Singleton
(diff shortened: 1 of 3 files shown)


chenlica:
@baiqiushi can you review and test this PR?

examples/twittermap/noah/src/main/scala/edu/uci/ics/cloudberry/noah/TwitterJSONTagToADM.scala
```diff
@@ -63,7 +63,7 @@ object TwitterJSONTagToADM {
 
     val thpool = Executors.newFixedThreadPool(threadNumber)
     implicit val ec: ExecutionContext = ExecutionContext.fromExecutor(thpool)
-
+
```
baiqiushi: Please avoid this unnecessary change.

examples/twittermap/web/app/controllers/TwitterMapApplication.scala
```diff
@@ -113,49 +107,23 @@ class TwitterMapApplication @Inject()(val wsClient: WSClient,
...
-        query.setResultType(resultType)
-        query.setGeoCode(centerLoc,radius,unit)
-        val tweetsResult = twitterAPI.search(query).getTweets
-        for(i <- 0 to tweetsResult.size()-1){
-          val status = tweetsResult.get(i)
-          if (status.isRetweet == false){
-            tweetArray = tweetArray :+ (Json.obj("id" -> status.getId.toString))
-          }
 
+        val url = "https://twitter.com/i/search/timeline?f=tweets&vertical=news&q="+queryWords+"&src=typd"
```
baiqiushi: Since this new way could be deprecated really easily in the future due to many reasons (e.g. Twitter updated the timeline page structure or limit the connection within OAth verification, etc. ), is it a better idea to keep the previous version as well by adding a flag in the configuration file to turn this hack version ON or OFF?
Just my thoughts, not sure which one is better. We may need the input from @chenlica , what do you think?

baiqiushi: Anyway, it would always be better if you add one line comment here to describe the main logic of the code block.

PR #686, opened 2019-05-06
[Frontend] Fix timebar blink line

AC-Unicorn:
### Problem description 
There will be a line blinking on the sidebar when a user searches for a keyword.
this issue only triggered by a popular keyword. 
here we have an example "watergate"
image
image

### The cause of the issue 
When the user issues a keyword query.
the time series will receive two results(There will be an empty result been received when
the controller is initializing) subsequently in a very short period. 
the first result will be small, and the second one will be large, 
image
In the figure above we can see the small result only has a length of 36, but the large one has a length of 4366. The small result will be first drawn as a line, then after a very short time, the entire diagram will be drawn using the large result. The line generated by the smaller result will be squeezed since the scale on Y-axis will soon  increase from a small number to a large number(in this example 0-50 to 0-1400)

### Solution 
To get rid of this line is simple, we can skip the first small result then directly draw the large result.


chenlica:
Why did the server send two results?  Can we get rid of the first, small result?

AC-Unicorn:
> Why did the server send two results? Can we get rid of the first, small result?

I am looking into this problem right now, this maybe intricate, because of time series also rely on cache result

examples/twittermap/web/public/javascripts/timeseries/controllers.js
```diff
@@ -80,6 +84,16 @@ angular.module('cloudberry.timeseries', ['cloudberry.common'])
       }
     };
 
+    function eventHandler(){
+      //reset time series state
+      $scope.timeseriesState = 1;
```
baiqiushi: Why the reset value is 1 but the initial value is 0?

AC-Unicorn: When `resultArray` is created, watch function will be triggered, even if we didn't actually receive a new result. So I want to distinguish the initializing state and the first-time receive query result state.

PR #690, opened 2019-05-14
[Graph Viz] Change query to be on keyword

sadeemsaleh:
ezgif com-video-to-gif
Previously the search was done based on the number of returned edges. Now, the search is based on a keyword.
Also removed unnecessary files that were automatically included in the hello world tutorial of play framework


chenlica:
Since it's a research prototype, we can review and merge this PR quickly.

PR #691, opened 2019-05-14
[Graph-Viz]Add a map in mapboxgl to the front-end

vichgy:
Add a map as the background of the interface.


chenlica:
For public shared js libraries, do we need to include them in our repo or we can just cite them via a URL?

Also the subject of this PR needs to be more informative.  @sadeemsaleh please provide the guidance to @vichgy

vichgy:
The original public shared js libraries, there are already files like deckgl.js and mapboxgl.min.js. Because the demo I followed used another version of deckgl and mapbox so that I just update the content in the file. We can also just import the js file from the URL. Which way do we prefer?

chenlica:
Can we have a URL with the version number to make it stable?

In the subject of the PR, please add [Graph-Viz] as a prefix.

vichgy:
So maybe we can add the URL and the version as a comment?

chenlica:
We don't want to include external libraries unnecessarily.  Can you find a permanent URL with a right version?

sadeemsaleh:
conflicts resolved in #703

PR #692, opened 2019-05-17
[Fontend] Fix number of scale of timebar

AC-Unicorn:
## Issue background
There is a minor usability issue existing in our current time series. The number of ticks(scale on Y-axis) some times are too many, then the scale maybe overlaps on each other.
image
Like the figure is shown, sometimes the Y-axis is too crowded 

## solution
Now we set the number of scales to a fixed number, for example 4.
image


AC-Unicorn:
> So this fixed scale number means there will always be 4 numbers appear on the y axis, but the real value of each number is still adjustable based on the values of the line chart, correct?

Yes the value will be automatically adjusted respect to the total count.

baiqiushi:
@AC-Unicorn Please resolve the conflicts and merge this PR.

PR #697, opened 2019-06-01
[graph-viz] Point Implementation and Remove Same Edges

yuboding:
1. Implemented searching with query in current window range. Search will now return edges that base on the query provided and have from/to points inside the current window range. 
2. Removed "Same" edges that have the same from coordinates and to coordinates, which appears the same in the Deck.gl demo (overlap). Edges removed could have different information.


sadeemsaleh:
conflicts resolved in #703

PR #700, opened 2019-06-20
Delta heatmap & heatmap cache

XinyueHan7:
1. Enable delta option to retrieve heatmap results by batches.
2. Add cache on heatmap to store and reuse returned data.


chenlica:
@AC-Unicorn @baiqiushi please review and finish this PR.  Thanks.

examples/twittermap/web/public/javascripts/heatmapcache/services.js
```diff
@@ -0,0 +1,50 @@
...
+      var day = new Date(store[j]["create_at"]);
+      if (day >= timeInterval.start && day <= timeInterval.end) {
+        resultArray.push(store[j]);
+      }
+    }
+    return resultArray;
+  };
+
+  /**
+   * Updates the store with map result each time the middleware responds to json request.
```
baiqiushi: If each time gets a partial result, we `putValues` the results into cache, the logic of updating `currentTimeRange` is not correct. It should be extended instead of assigned the `timeInterval` of current partial results.

examples/twittermap/web/public/javascripts/map/heatmap.js
```diff
@@ -76,25 +77,47 @@ angular.module("cloudberry.map")
...
+        else {
+          cloudberry.heatmapMapResult = heatMapResultCache.getValues(cloudberry.parameters.timeInterval);
+        }
       }
+    
       // Partial time series cache hit case
       else {
+        var heatTimeJson = queryUtil.getTimeBarRequest(cloudberry.parameters, $scope.geoIdsNotInTimeSeriesCache);
+      
+        cloudberryClient.send(heatJson, function(id, resultSet, resultTimeInterval) {
```
baiqiushi: Why here the checking logic of `heatMapResultCache.cacheIsDone` in the above case is gone? This `else` case is for when `$scope.geoIdsNotInTimeSeriesCache.length != 0`, not for `heatMapResultCache`.

examples/twittermap/web/public/javascripts/map/heatmap.js
```diff
@@ -76,25 +77,47 @@ angular.module("cloudberry.map")
...
+          if(angular.isArray(resultSet)) {
+            cloudberry.commonTweetResult = resultSet[0].slice(0, queryUtil.defaultSamplingSize - 1);
+            cloudberry.heatmapMapResult = resultSet[0];
+            heatMapResultCache.putValues(cloudberry.heatmapMapResult, cloudberry.parameters.timeInterval)
```
baiqiushi: Here it proves the above conjecture, that we call `putValues` every time we get results from Cloudberry, and since we enabled the `delta` feature, then each time the results coming from Cloudberry have disjoint time range.

examples/twittermap/web/public/javascripts/heatmapcache/services.js
sophiewu6: I have changed it to the partial time range, but I realized there is a small problem with it. For example, since keyword 'Trump' has a large amount of data, it will only fetch data for one to two months. In this case, it will never be considered as 'complete heatmap cache hit case' because the entire time range is one year, but the cache time range is only one or two months.

baiqiushi: Sorry, I didn't follow. Can you elaborate more?

baiqiushi:
No needed anymore.

PR #722, opened 2019-09-01
[Graph-viz] Incremental Hierarchical Clustering and FDEB

SustechJoy:
Mark all revised code to be modified with TODO comment.


graph-viz/app/algorithms/PointCluster.java
```diff
@@ -0,0 +1,211 @@
...
+// TODO try to make the cluster not imaginary to make the incremental bundling easier
+public class PointCluster {
+
+    private int minZoom;
+    private int maxZoom;
+    // TODO understand radius
+    private double radius = 80;
+    // TODO explain and add comment for extent
+    private double extent = 512;
+    private KdTree[] trees;
```
sadeemsaleh: add a comment explaining there is a kdTree for every zoom level

graph-viz/app/controllers/GraphController.java
```diff
@@ -15,73 +23,490 @@
...
+        ArrayList<Cluster> points = new ArrayList<>();
+        while (resultSet.next()) {
+            resultSetSize++;
+            double fromLongitude = resultSet.getDouble("from_longitude");
+            double fromLatitude = resultSet.getDouble("from_latitude");
+            double toLongitude = resultSet.getDouble("to_longitude");
+            double toLatitude = resultSet.getDouble("to_latitude");
+            Edge currentEdge = new Edge(fromLatitude, fromLongitude, toLatitude, toLongitude);
+            if (edgeSet.contains(currentEdge)) continue;
+            points.add(new Cluster(fromLongitude, fromLatitude));
```
sadeemsaleh: seems you are creating a cluster for every point initially

graph-viz/app/controllers/GraphController.java
```diff
@@ -15,73 +23,490 @@
...
+            lineNode.put("width", entry.getValue());
+            arrayNode.add(lineNode);
+        }
+        json = arrayNode.toString();
+        objectNode.put("data", json);
+        objectNode.put("isolatedEdgesCnt", edges.size());
+    }
+
+    private double sqDistance(double x1, double y1, double x2, double y2) {
+        return Math.pow(x1 - x2, 2) + Math.pow(y1 - y2, 2);
```
sadeemsaleh: Is this square distance? then why not do Math.sqrt? Also why not move the function distanceTo from Cluster class to Point class and use it here since the points would be defined from Point class

graph-viz/app/controllers/GraphController.java
```diff
@@ -15,73 +23,490 @@
...
+        if (!endCalendar.before(PropertiesUtil.lastDateCalender)) {
+            System.out.println(finished + end);
+            objectNode.put("flag", finished);
+        } else {
+            System.out.println(unfinished + end);
+            objectNode.put("flag", unfinished);
+        }
+    }
+
+    public void cluster(double lowerLongitude, double upperLongitude, double lowerLatitude, double upperLatitude, int clustering, int clusteringAlgorithm, String timestamp, int zoom) {
```
sadeemsaleh: too many repeated lines in here. move all similar logic to one function

graph-viz/app/controllers/GraphController.java
```diff
@@ -15,73 +23,490 @@
...
+                }
+                json = arrayNode.toString();
+            }
+        } else if (clusteringAlgorithm == 1) {
+            if (iKmeans != null) {
+                if (clustering == 0) {
+                    ArrayNode arrayNode = objectMapper.createArrayNode();
+                    pointsCnt = iKmeans.getPointsCnt();
+                    clustersCnt = pointsCnt;
+                    for (int i = 0; i < iKmeans.getK(); i++) {
```
sadeemsaleh: why is there a nested loop?

graph-viz/app/controllers/GraphController.java
```diff
@@ -15,73 +23,490 @@
...
+                }
+                externalEdgeSet.add(edge);
+            }
+        }
+    }
+
+
+    private void getKmeansEdges(int zoom, int bundling, int clustering, ObjectNode objectNode, HashMap<models.Point, Integer> parents, ArrayList<double[]> center) {
+        HashMap<Edge, Integer> edges = new HashMap<>();
+        if (clustering == 0) {
```
sadeemsaleh: I think there won't be such a case

graph-viz/app/algorithms/ForceBundling.java
```diff
@@ -0,0 +1,409 @@
...
+        return p.getX() * q.getX() + p.getY() * q.getY();
+    }
+
+    /**
+     * Calculates the vector format of edge.
+     * @param e current edge.
+     * @return corresponding vector.
+     */
+    private Point edgeAsVector(EdgeVector e) {
+        return new Point(dataNodes.get(e.getTargetNodeInd()).getX() - dataNodes.get(e.getSourceNodeInd()).getX(), dataNodes.get(e.getTargetNodeInd()).getY() - dataNodes.get(e.getSourceNodeInd()).getY());
```
sadeemsaleh: Once you use Point Edge class instead of an EdgeVector and you would also have a point attribute in the Edge class, then you would not need

graph-viz/app/algorithms/ForceBundling.java
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    private double edgeLength(EdgeVector e) {
```
sadeemsaleh: Use the function from Point class between two points

graph-viz/app/algorithms/ForceBundling.java
```diff
@@ -0,0 +1,409 @@
...
+
+    /**
+     * Calculates distance between two points,
+     * the distance of two edges is given by the distance
+     * between the middle points of these two edges.
+     * @param p source node.
+     * @param q target node.
+     * @return distance of two points.
+     */
+    private double euclideanDistance(Point p, Point q) {
```
sadeemsaleh: Use from the function which will be added to Point class

graph-viz/app/algorithms/TreeCut.java
```diff
@@ -0,0 +1,240 @@
...
+     * @param upperLongitude  upperLongitude of user screen
+     * @param lowerLatitude   lowerLatitude of user screen
+     * @param upperLatitude   upperLatitude of user screen
+     * @param zoom            zoom level of user screen
+     * @param edges           edge set to be returned
+     * @param externalEdgeSet edge set with only one node inside screen
+     * @param externalCluster outside cluster corresponding to edge set with only one node inside screen
+     * @param internalCluster inside screen clusters
+     */
+    public void treeCut(PointCluster pointCluster, double lowerLongitude,
```
sadeemsaleh: give a meaningful name such as init. and add constructors in the class

graph-viz/app/treeCut/TreeCut.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * add the mapping results to the externalCLusterMap
+     * @param externalClusterMap cluster in the final result each external cluster mapped to
+     * @param entry external hierarchy map entry
+     * @param elevateLevel the level that the clusters to be elevated
+     */
+    private void updateExternalClusterMap(HashMap<Cluster, Cluster> externalClusterMap, Map.Entry<Cluster, ArrayList<Cluster>> entry, int elevateLevel) {
+        Cluster mappedToCluster;
+        for (Cluster c : entry.getValue()) {
+            mappedToCluster = c;
```
sadeemsaleh: mappedToCluster rename it. it should also point to mappedToCluster.parent;

graph-viz/app/treeCut/TreeCut.java
```diff
@@ -0,0 +1,238 @@
...
+        HashSet<Cluster> internalHierarchy = new HashSet<>();
+        // initialize the hierarchy
+        addExternalClusterHierarchy(externalCluster, externalHierarchy, externalClusterMap);
+        addInternalClusterHierarchy(internalCluster, internalHierarchy);
+        // TODO change breaking clause
+        while (externalHierarchy.size() != 0) {
+            // The ancestors has conflict
+            ArrayList<Cluster> removeAncestors = new ArrayList<>();
+            // Find clusters has common ancestor with internal clusters at this level
+            for (Map.Entry<Cluster, ArrayList<Cluster>> clusterEntry : externalHierarchy.entrySet()) {
```
sadeemsaleh: Instead of using removeAncestor to remove unwanted items, use stream.map to delete

graph-viz/app/treeCut/TreeCut.java
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Elevates the hierarchy of external clusters
+     *
+     * @param externalHierarchy  external cluster map to the ancestor
+     * @param externalClusterMap cluster in the final result each external cluster mapped to
+     * @return elevated cluster
+     */
+    private HashMap<Cluster, ArrayList<Cluster>> elevateExternalHierarchy(HashMap<Cluster, ArrayList<Cluster>> externalHierarchy, HashMap<Cluster, Cluster> externalClusterMap) {
+        HashMap<Cluster, ArrayList<Cluster>> tempExternalHierarchy = new HashMap<>();
+        for (Map.Entry<Cluster, ArrayList<Cluster>> entry : externalHierarchy.entrySet()) {
```
sadeemsaleh: Try not to create a new instance. do the same for internal

graph-viz/app/treeCut/TreeCut.java
```diff
@@ -0,0 +1,238 @@
...
+            }
+            // remove all clusters stop at this level
+            for (Cluster c : removeAncestors) {
+                externalHierarchy.remove(c);
+            }
+            // elevate all remaining clusters to a higher level
+            externalHierarchy = elevateExternalHierarchy(externalHierarchy, externalClusterMap);
+            internalHierarchy = elevateInternalHierarchy(internalHierarchy);
+        }
+        for (Edge edge : externalEdgeSet) {
```
sadeemsaleh: might be better to loop inside the method

graph-viz/app/treeCut/TreeCut.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Not applying tree cut algorithm
+     *
+     * @param pointCluster    hierarchical structure from HGC algorithm
+     * @param zoom            zoom level of user screen
+     * @param edges           edge set to be returned
+     * @param externalEdgeSet edge set with only one node inside screen
+     */
+    public void nonTreeCut(PointCluster pointCluster, int zoom, HashMap<Edge, Integer> edges, HashSet<Edge> externalEdgeSet) {
```
sadeemsaleh: No need for this method. We can use the one from graphController

graph-viz/app/controllers/IncrementalQuery.java
```diff
@@ -0,0 +1,65 @@
...
+import utils.PropertiesUtil;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.util.Calendar;
+
+/**
+ * Implement the incremental database query
+ */
+public class IncrementalQuery {
```
sadeemsaleh: This class does not extend Controller class so it is not really a controller maybe move it to some other package

graph-viz/app/controllers/WebSocketController.java
```diff
@@ -0,0 +1,34 @@
...
+    @Inject
+    public WebSocketController(ActorSystem actorSystem, Materializer materializer) {
+        this.actorSystem = actorSystem;
+        this.materializer = materializer;
+    }
+
+    /**
+     * Forwards requests to the bundle actor.
+     */
+    public WebSocket socket() {
```
sadeemsaleh: Where is this used?

graph-viz/test/controllers/EdgeTest.java
```diff
@@ -1,12 +1,14 @@
 package controllers;
+import models.Edge;
+
 import java.util.HashSet;
 import java.util.Set;
 public class EdgeTest{
     public static void main(String[] args) {
```
sadeemsaleh: This class is not testing anything significant so might be better to delete it

graph-viz/app/models/Edge.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Objects;
+
+public class Edge {
+
+    // from Point of edge
+    private Point from;
+    // to Point of edge
+    private Point to;
+
+    public Edge(double fromX, double fromY, double toX, double toY) {
```
sadeemsaleh: Since we are using Point and not double, minimize the confusion by actually passing Point not double as the parameters

graph-viz/app/controllers/WebSocketController.java
SustechJoy: In the conf/routes file.

graph-viz/app/connection/Response.java
```diff
@@ -0,0 +1,149 @@
...
+    private String data;
+    // the number of edges on the screen that are not bundled
+    private int isolatedEdgesCnt;
+    // the end date of the current slice
+    private String date;
+    // the number of points (or points in clusters) on the screen
+    private int pointsCnt;
+    // the number of clusters on the screen
+    private int clustersCnt;
+
```
sadeemsaleh: This class has no constructor? are you setting every variable individually?

graph-viz/graph-viz.iml
```diff
@@ -0,0 +1,204 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sadeemsaleh: What is this file for?

PR #725, opened 2019-10-31
[Elasticsearch Adapter] Support aggregation functions in GROUP BY statement

dayuebai:


cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/model/impl/ElasticsearchGenerator.scala
```diff
@@ -440,15 +471,23 @@ class ElasticsearchGenerator extends IQLGenerator {
               groupStr.append(s""" "terms": {"field": "${field}", "size": $MAX_RESULT_WINDOW} """.stripMargin)
             }
           }
-          // TODO: Parsing response of Sum, Avg, Max, Min aggregation query will be implemented in later PR.
+          case GeoCellTenth => ??? // Elasticsearch adapter does not support geo cell functions.
```
baiqiushi: Maybe it's better to add one sentence to indicate the reason? Because Elasticsearch does not support geo-location data binning operations? Or because it's too complicated to implement.

PR #731, opened 2020-02-21
[Fix bug] Fix tweet url composition with user.id as part of the link.

baiqiushi:
Fix the bug of tweet url with user.id as part of the url instead of previous user.name which is not an identifier of a user;


chenlica:
LGTM!

PR #754, opened 2020-03-22
[CoronavirusTwitterMap] Show case numbers data in popup window of countmap

baiqiushi:
#754 
image


chenlica:
Looks great!  Here are some comments:

- Show the case number chart at the top;
- We can show the three numbers in a better way.  Maybe a tiny table? 
- Add the population number;
- Use the "X,XXX" format to show those number, such as "1364 -> 1,364"; Do the same for the y-axes;
- When I clicked California, part of the popup window is outside the screen. Please adjust the initial location.

PR #758, opened 2020-03-23
[CoronavirusTwitterMap] Fix popup window minor bug

baiqiushi:


baiqiushi:
Include other branch, close and re-submit.

PR #767, opened 2020-03-24
Enable Google Analytics

Yicong-Huang:
add Global site tag (gtag.js) for Google Analytics


examples/coronavirustwittermap/web/app/views/twittermap/main.scala.html
```diff
@@ -8,6 +10,20 @@
 
 <head>
   <title>@title</title>
+
```
baiqiushi: Let's make to enable/disable google analytics configurable in application.conf first. Then any machine, user related parameters should also be configurable.

Yicong-Huang: How about the current setup? 

- in conf needs to specify a `gtag.id` variable
- by default or omitted, this id is empty ("")
- the google script will be load and run anyway. but without a proper gtagId, it will not do anything. 

So in conclusion:
- To enable Google Anayalitics, add a valid `gtag.id` in conf.
- To disable Google Analytics, omit the `gtag.id` in conf.

baiqiushi: OK. As long as if `gtag.id` is omitted in application.conf, this piece of code will not do anything harm to the frontend.

Issue #676, opened 2019-04-09
AsterixDB scripts need to be updated


luochen01:
AsterixDB recently introduced a patch https://asterix-gerrit.ics.uci.edu/#/c/3267/ that no longer supports using the request body as statement. Instead, a parameter called "statement" must be provided to define a set of queries to be executed. As a results, many scripts used by Cloudberry will not work and must be fixed when upgrading AsterixDB.

The following changes can be used to fix the scripts:

1. removes -XPOST (as POST is implicit for curl's --data options)
2. uses URL encoding instead of binary encoding for the content, and
3. provides the parameter name statement on the curl command line instead if providing it in the body.

An example is as follows:
```
cat <<EOF | curl --data-urlencode statement@- http://localhost:19002/query/service 
drop dataverse twitter if exists;
create dataverse twitter if not exists;
use twitter;
...
EOF
```

Reference: https://issues.apache.org/jira/browse/ASTERIXDB-2543

baiqiushi:
#771

PR #771, opened 2020-03-25
Update AsterixDB API usage

parshimers:
The AsterixDB API was cleaned up in https://issues.apache.org/jira/browse/ASTERIXDB-2659. Some of the scripts used these old methods of sending queries that were removed. A parameter also had the set of valid paramters changed.


baiqiushi:
#676

Issue #761, opened 2020-03-23
Add a query-caching module on TwitterMap server


chenlica:
For each query sent from the UI, cache the query results on the TwitterMap server.  Invalidate it after some time (e.g., 5 mins). This method can reduce the number of queries to Cloudberry.

Yicong-Huang:
@chenlica just saw this now. We did not discuss this today, is it still needed for now?

baiqiushi:
Done

PR #783, opened 2020-03-31
[CoronaVirus] add caching module issue #761

sadeemsaleh:
- when a query is in the cache, the result would be returned from the cache after checking its freshness
- the cache is cleaned whenever we hit the max threshold of some number of queries.
- every thread would have its local copy to maintain the cache until it receives the done.
- once the done message is received, the local cache would push the changes to the central cache.
- whenever two threads try to add to the central cache, it would take the latest timestamp.
- the timestamp is created upon request of the query not the result of the query.

### DIFF
--- a/examples/coronavirustwittermap/project/dependencies.scala
+++ b/examples/coronavirustwittermap/project/dependencies.scala
@@ -51,7 +51,9 @@ object Dependencies {
     "org.twitter4j" % "twitter4j-stream" % twitter4jVersion,
     "org.twitter4j" % "twitter4j-core" % twitter4jVersion,
     // case number module
-    "org.webjars.bower" % "jquery-csv" % "0.8.3"
+    "org.webjars.bower" % "jquery-csv" % "0.8.3",
+    // handle big integers in javascript to support safari
+    "org.webjars.npm" % "jsbi" % "3.1.1"
 
   ) ++ testDeps
 }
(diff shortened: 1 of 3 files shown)


Yicong-Huang:
resolves #761

cloudberry/zion/src/main/scala/edu/uci/ics/cloudberry/zion/actor/RESTSolver.scala
```diff
@@ -33,5 +33,4 @@ class RESTSolver(val dataManager: ActorRef,
       }
   }
 
-}
-
+}
```
baiqiushi: Let's avoid unnecessary changes.

examples/coronavirustwittermap/web/app/websocket/TwitterMapServerToCloudBerrySocket.scala
```diff
@@ -68,24 +84,24 @@ class TwitterMapServerToCloudBerrySocket(out: ActorRef, config: Configuration) {
                 /**
                   * 1) pinMapResult response from Cloudberry has format as follows,
                   * { id: "defaultID",
-                  *   category: "pinMapResult",
-                  *   value: [[
-                  *     { id: 1236910669389131776, place.bounding_box: [[-120.582586,46.958017], [-120.49726,47.028542]] },
-                  *     { id: 1236509727560880128, coordinate: [-121.944442,37.325656], place.bounding_box: [[...]]},
-                  *     ... ...
-                  *   ]]
+                  * category: "pinMapResult",
```
baiqiushi: Let's avoid this unnecessary format changes, and also this indent is useful for readers.

sadeemsaleh: Sorry, I didn’t change this part must have been due to reformat code option

PR #789, opened 2020-04-03
Fix the popup window chart bug

sophiewu6:
Screen Shot 2020-04-03 at 11 35 37 AM

Currently, the chart shows wrong trend. This PR aims to fix this bug.

After:
Screen Shot 2020-04-03 at 11 37 41 AM


Yicong-Huang:
what was the issue?

PR #791, opened 2020-04-05
[Backend] Fix Cache Replacement Logic

sadeemsaleh:
fix a bug in replacing the cache which caused memory overflow


examples/coronavirustwittermap/web/conf/application.conf
```diff
@@ -200,4 +200,4 @@ gtag.id = ""
 
 # Server Cache module parameters
 # How many minutes at most the cache is valid. (Default: 10)
```
baiqiushi, line 202:
```suggestion
# How many milliseconds at most the cache is valid. (Default: 600000)
```

PR #795, opened 2020-04-08
[Frontend] Mobile UI Improvements

JosephZheng1998:
- Fixed Count Map bug where popup window does not close when switched to other maps
- Implemented mobile-friendly Hot Topics
- Fixed the PinMap clicking issue due to the Polygon Layer


examples/coronavirustwittermap/web/public/javascripts/map/pinmap.js
```diff
@@ -168,15 +167,18 @@ angular.module("cloudberry.map")
             }
             $scope.previousZoomLevel = event.level;
             sendPinmapTimeQuery();
+
+            $scope.deletePolygonLayers();
```
baiqiushi, line 171: I think these delete calls made us can not find the bug when they will be added again. Since we know where they are added, let's remove these delete calls. Only call the delete when we switch to pinmap from countmap.

PR #807, opened 2020-06-17
Data Tools with Twitter Ingestion Server, Twitter GeoTagger and AsterixDB Ingestion Server.

baiqiushi:
## Data Tools
**Data Tools** is a new module consisting of 3 components that serve the data preparation of the TwitterMap application.
### Twitter Ingestion Server
**Twitter Ingestion Server** is a daemon service that can ingest real-time tweets from [Twitter Filter Stream API](https://developer.twitter.com/en/docs/tweets/filter-realtime/api-reference/post-statuses-filter) into local gzip files in a daily rotation manner.
It is also a light-weight HTTP server with 3 endpoints:
 - `/stats` - HTTP GET endpoint that returns current ingestion status information in JSON format.
 - `/proxy` - WebSocket endpoint that pushes real-time tweets to any client in connection.
 - `/` - HTTP GET endpoint that returns an `index.html` as an example page demonstrating the usage of the above two endpoints.

### Twitter GeoTagger
**Twitter GeoTagger** is Java program to geoTag Twitter JSON with `{stateID, stateName, countyID, countyName, cityID, cityName}`. 
It has 2 modes, 
 - 1) in API mode, it provides a function `tagOneTweet` that can be called from other programs; 
 - 2) in process mode, it provides a main function that can be started as a JVM process to geotag tweets in shell console pipeline.

### AsterixDB Ingestion Server
TBD.

### DIFF
--- a/examples/twittermap/build.sbt
+++ b/examples/twittermap/build.sbt
@@ -45,4 +45,4 @@ lazy val datatools = (project in file("datatools")).
   settings(Commons.settings: _*).
   settings(
     libraryDependencies ++= datatoolsDependencies
-  )
+  ).dependsOn(gnosis, util)
--- a/examples/twittermap/build.sbt
+++ b/examples/twittermap/build.sbt
@@ -40,3 +40,9 @@ lazy val guardian = (project in file("guardian")).
   settings(
     libraryDependencies ++= guardianDependencies
   )
+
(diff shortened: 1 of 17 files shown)


sadeemsaleh:
NOT ready to be merged

PR #814, opened 2020-10-22
Upgrade the twitter4j version to 4.0.7 to support full_text

baiqiushi:
In the `noah` sub-project, twitter json parser is using twitter4j for both geotagging and json to ADM transformation;
The current version (4.0.3) of twitter4j does not support `full_text` attribute and the `text` atrribute will be the truncated result from `full_text`;
After upgrading to version (4.0.7), twitter4j will fill the `text` attribute with the content from `full_text` by default, so that the ingested tweets have the orignal text without truncating.


sadeemsaleh:
ready to merge, waiting for check

PR #818, opened 2021-01-29
[Frontend] Fix the time interval bug in time series bar

baiqiushi:
Fix the problem of a small number was hardcoded as the defaultNonSamplingDayRange for queries without a SamplingDayRange setup;


baiqiushi:
Solved #821

